PR #10, opened 2022-10-21
Added PinModel + DigitalOutput + DigitalInput + Creation of ST-LIB Library

StefanCostea:
·Added PinModel (Used to describe, initialize and register PINs)
·Added DigitalOutput Service (Used to set binary voltage values on a PIN) 
·Added DigitalInput Service (Used to get binary voltage values from a PIN)
·Created ST-LIB as a STM32Cube Static Library
·Fixed Issue where declaration and implementation couldn't be on separate files
·Fixed Issue with PinModel that disabled STLINK
·Performed test to check functionality:
image
Pin PB0 had a led connected which toggled each 500ms how it should be, while PC3 was configured as an input and read PinState::ON when 3,3V were connected to it and PinState::OFF when GND was connected to it


PabloGonzalezSegarra:
No tienes el .gitignore y se ha subido la carpeta Debug. ¿Qué ha pasado con la carpeta drivers que da 2000 conflictos?. Esto va a ser un empastre.

PabloGonzalezSegarra (approved):

PR #45, opened 2022-11-20
SPI service

PabloGonzalezSegarra:
# **Pull request of the SPI service**

The service is completely finished, but due to technical problems with the workshop material, the functions `SPI::is_busy() `and `SPI::has_next_packet` have not yet been able to be tested. The rest have been tested using _Analog Discovery 2_ checking that messages are sent and received correctly.

## Operation of the SPIPacket

The `SPIPacket` class is a simple class that manages the packets to be sent or received by SPI. The only thing it does is store any type of data either single or an array in a byte array. Therefore the user of this class does not need to know in which format they should be sent and facilitates the use of the `SPI` class. Its use is explained below:

code

## Operation of the SPI service

The operation of the service is quite simple. It should be registered like all services, with small changes mentioned later. Once registered, SPI packets can be sent using the `SPIPacket` class. The packet can be of the desired size, so you can send as many bytes as you want. Once the command `SPI::transmit_next_packet(uint8_t id, SPIPacket& packet)` has been sent, before sending another one, it must be checked that the SPI is not occupying using the method `SPI::is_busy(uint8_t)` to avoid that the method returns `false`. Similarly, when sending an order to receive a package using `SPI::receive_next_packet(uint8_t id, SPIPacket& packet)`, it must be checked that the value of the package is already correct using `SPI::has_next_packet(uint8_t id)` before using it. This is because the SPI is configured in interrupts mode, this allows the microcontroller to continue doing other things while a message is being sent or received. 

Below is an example of using the class, although in this example in order to keep it simple it has not been done, to use the SPI it must be done through small state machines to prevent the microcontroller from being blocked while waiting for it sending or receiving a package. The example emulates the reading of an ADC sensor.

code

## What's new with respect to other services 

To further abstract the hardware to HALAL users some changes have been added. These changes not only abstract the use of HALAL but also simplify the code, make it more robust and save maps and checks. To achieve this, a private `SPI::Peripheral` structure has been added within the service class and later all possible peripherals are declared within this class. In this way only the available peripherals can be used. This code is perfectly self-generating, which makes it very modular.

code

This structure contains all the information regarding the SPI peripheral that is being used, therefore only one map is necessary to store all the registered SPIs. The structure is instantiated in the file `.cpp` and is done as follows:

code

Another change made is that the normal `map` maps have been changed to 'unordered_map' as they are more efficient in this case and due to the changes mentioned above it is not necessary to add anything extra.

Declaration in the `.hpp:`

code

Usage in the `.cpp`:

code

These changes allow us to avoid checks like the one we had in the registers, checking that the pins (or anything else) belonged to a real peripheral. Now this is not necessary because only existing SPI peripherals can be passed to it since only the preset peripherals can be used. Register method with the new changes:

code

Using the new register method:

code

## Tests

Except for the methods mentioned at the beginning, the rest have been tested individually using the _Analog Discovery 2_  and their operation has been correct. To ensure that all methods and classes are integrated correctly, the following test has been designed.

### Sensor emulator test

> Important: This test has not yet been carried out for the reason stated at the beginning. When it is done, you will be notified.

Two _nucleo_ have been programmed, one as a master pretending to be a common board that reads data from an ADC and sends it to the computer, in this case also using SPI and reading it with _Analog Discovery 2_, and another _nucleo_ that pretends to be an ADC sensor that upon receiving the correct order, it sends the last value read.

The ADC sensor has been programmed as _slave_, this board waits to receive a packet and depending on it sends one value or another. To keep the test simple, only one value has an answer, the rest are considered invalid. The code is the following:

code

The board has been programmed as _master_. This board sends the correct order and waits for the arrival of the information, then sends it to the computer. The code is the following:

code

Remember that although it has not been done in the examples provided and tests carried out, it is very important to use state machines when using SPI to avoid blocking the microcontroller.


StefanCostea:
Input for service should be at least a primitive type, try masking SPIPeripheral Class with an Enum or something, this is needed so that mocks are possible to develop

alejandrogonzalvo (changes requested):

PabloGonzalezSegarra:
Ya he pasado todos los tests de forma satisfactoria.

alejandrogonzalvo (approved):

StefanCostea (approved):

PR #66, opened 2022-11-30
PWM class 

alejandrogonzalvo:
# PWM
This PR implements the PWM class on top of the PWM service, that allows to create PWM instances. It will be used for the half bridges and the linear induction motor (LIM).

## Usage
``` cpp
int main(void)
{
  /* USER CODE BEGIN 1 */
	PWM pwm1 = PWM(PE6);
  /* USER CODE END 1 */

  // Inits and HAL code...
  
/* USER CODE BEGIN 2 */

  pwm1.turn_on();
  pwm1.set_duty_cycle(70);

  /* USER CODE END 2 */

  /* Infinite loop */
  /* USER CODE BEGIN WHILE */
  while (1) {}
}
```

## Implementation
``` cpp
PWM::PWM(Pin pin) : pin(pin) {
	optional<uint8_t> try_id = PWM::inscribe(pin);
	if (not try_id) {
		//TODO: error handler
	}

	id = try_id.value();
}

void PWM::turn_on() {
	PWM::turn_on(id);
}
```

## Testing
I have checked that the pwm is generated correctly with waveforms and the analog device 2.
image
Download the test [here](https://github.com/alejandrogonzalvo/ST-LIB-Tests/tree/main/STLIB-LOW_PWM)

PR #72, opened 2022-12-02
Features/digital output

alejandrogonzalvo:
# Digital Output
This PR implements the Digital Output class on top of the Digital Output service, that allows to create Digital Output instances. It will be used for the half bridges and the linear induction motor (LIM).

Works exactly the same as the PWM class, see [this PR](https://github.com/HyperloopUPV-H8/ST-LIB/pull/66) for more info.


PabloGonzalezSegarra:
Para que queremos esta clase? La idea de los servicios es que no tienes instancias. No comprendo en que caso esto tiene utilidad.

PR #81, opened 2022-12-09
Features/cordic

ricardochust:
CORDIC class done and tested

The class works in a very simple way, it doesn t even need any start method. If you want a sine, you just use the sine method and it does all the magic alone.
First it checks if its properly configurated, and if its not then configurates it. After that, the method just feeds the CORDIC until it has no more values to calculate on the in pointers, and fills the out pointers with the results. 

The calculations all work with int values, which represent different things for each method. It uses its own units, but if we work with these units on all calculations there shouldn t be any problem.

The units work as follows:

- angle: the max int value (2147483647) represents 180 degrees, or pi radians. After that it just wraps into the negatives. It seems like that would be troublesome, but it is made so overflow and underflow work on our favour, as if we add 2º to 180º we still get 182º (as it overflows to -2123622718, which represents 182º). In uint logic it is just 360 positive degrees no problem. 

- distances: very simple, from a point of reference, 2147483647 is the maximun positive and -2147483647 is the maximun negative. Point. Using very low values (like 15 000) may give less precise results, so do not fear using reasonably high numbers. In uint logic when we get over the maximun value it just appears on the other side, so beware with overflows on these units. 
 
- cos and sine results: as simple as distances. 2147483647 is one (the max value possible) and -2147483647 is minus one (the minimun value possible). Very simple. 

The tests were quite rudimentary, just testing some values on the correct units, calculated them 5 000 000 times and see how much times they take and if they were correct. Some values tested were 0, pi/3, pi/2, pi -pi/2 for cosine and sine and (1 0) (1 1) (0 1) (0.3 1) (-1 0) (-0.5 -1) for the modulus and phase (traduced to ints and put against arm.math with ints when possible).
They seemed to be at least twice as fast as arm.math and also have the advantage to be able to calculate sines and cosines or modulus and phase in just one take (which makes it potentially 4 times as fast)

If we wanted even more speed we could reduce the precision quite easily by changing the configuration step inside of each method. I though on doing a variable for precision to make it even easier to change, but it would need an OR calculation on each configuration and it is not like we are going to change that each week.

At last, the methods can work just as good with only one input on the pointers, you just need to make the size parameter one. If you will do multiple calculations it is obviously better to give all of them at once, but you just lose 2 comparations and one jump so its not gonna explode. You can even give a pointer to a 1000 matrix and tell it to just use the first 100, just put the size parameter to 100 and it will use the 100 first and then stop. (disclaimer: it wouldn t work if you give 100 matrix and put a 1000 in the size parameter, don t do that, it will go out of the pointer)


Inc/HALAL/Services/CORDIC/CORDIC.hpp
```diff
@@ -0,0 +1,45 @@
...
+ * Created on: 29 nov. 2022
+ * 		Author: Ricardo
+ */
+
+#pragma once
+
+#include "PinModel/Pin.hpp"
+//#ifdef HAL_CORDIC_MODULE_ENABLED
+
+	enum Operation_Computation{
```
PabloGonzalezSegarra, line 13: Porque este enum esta identado?

Inc/HALAL/Services/CORDIC/CORDIC.hpp
```diff
@@ -0,0 +1,45 @@
...
+	static void cos(int32_t *angle,int32_t *out,int size);
+	static void sin(int32_t *angle,int32_t *out,int size);
+	static void cos_and_sin(int32_t *angle, int32_t *cos_out, int32_t *sin_out, int size);
+	//vector values can go from (-2147483647 to 2147483647)
+	// vector (x,0) is phase 0
+	static void phase(int32_t *x, int32_t *y, int32_t *angle_out, int size);
+	//cannot give a value higher than 2147483392
+	static void modulus(int32_t *x, int32_t *y, int32_t *out, int size);
+	static void phase_and_modulus(int32_t *x, int32_t *y, int32_t *angle_out, int32_t *mod_out, int size);
+	static Operation_Computation mode;
```
PabloGonzalezSegarra, line 40: Los atributos primero que los metodos.

Src/HALAL/Services/CORDIC/CORDIC.cpp
```diff
@@ -0,0 +1,123 @@
...
+ */
+
+#include "CORDIC/CORDIC.hpp"
+
+
+Operation_Computation RotationComputer::mode = NONE;
+
+
+
+void RotationComputer::cos(int32_t *angle, int32_t *out, int size){
```
PabloGonzalezSegarra, line 15: Intenta usar `int32_t` en vez de `int` normal. Si no sabes que tamaño quieres puedes usar `size_t`.

Src/HALAL/Services/CORDIC/CORDIC.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+void RotationComputer::phase_and_modulus(int32_t *x, int32_t *y, int32_t *angle_out, int32_t *mod_out, int size){
+	if(RotationComputer::mode != PHASE_MODULUS){
+			RotationComputer::mode = PHASE_MODULUS;
+			MODIFY_REG(CORDIC -> CSR,
+						0x007F07FF,
+						0x00180052
+			);
+		}
+		for(int n = 0; n < size; n++){
```
PabloGonzalezSegarra, line 96: Porque este for esta identado?

Src/HALAL/Services/CORDIC/CORDIC.cpp
```diff
@@ -0,0 +1,123 @@
...
+	 Ready flag : 0x80000000
+
+	 */
+float RotationComputer::q31_to_radian_f32(uint32_t in){
+	return M_PI*ldexp((int32_t) in, -31);
+}
+
+int32_t RotationComputer::radian_f32_to_q31(double in){
+	return (int)roundf(scalbnf(fmaxf(fminf(in/M_PI,0.9995),-0.9995),31));
+}
```
PabloGonzalezSegarra, line 122: Tienes que documentar muy bien con doxygen la clase. Una entrada en la wiki tampoco estaria mal. Es muy dificil entender el código. Además como he puesto antes usa #define para los numeros mágicos y explicalos.

PabloGonzalezSegarra (changes requested):
Review mia y de Stefan.

Inc/HALAL/Services/CORDIC/CORDIC.hpp
```diff
@@ -0,0 +1,144 @@
...
+	static void phase(int32_t *x, int32_t *y, int32_t *angle_out, int size);
+
+
+	/*
+	 * @brief modulus function. Calculates the modulus of the vector from (0,0) to (x,y). If the modulus result where to be higher than 2147483392 (~0.995) it will give aberrant result instead
+	 * @param x. The array of values x of each vector with size "size". input.
+	 * @param y. The array of values y of each vector with size "size". input.
+	 * @param out. The array where the functions puts its output, cannot be higher than 2147483392. output.
+	 *
+	 * the modulus function is a bit more fragile than the other functions.
```
PabloGonzalezSegarra, line 110: Empieza las frases en mayusculas.

PabloGonzalezSegarra (approved):

StefanCostea (approved):

PR #97, opened 2022-12-14
Features/math

ricardochust:
New class math

Receives int numbers that represent what is convenient for the calculations and outputs the result of such. 

Angle: goes from [-pi,pi] across all ints values. So MAX_INT would be pi.
Coordinates: goes from [-limit_space,limit_space] on both directions. MAX_INT would fall on the limit of the space, and one more would overflow into the -limit_space coordinate.
Unitary: goes from [-1,1], used as returns of cos and sine. 
tg: Units used on tg, goes from [-65536,65535] and has 16 bits for decimals. If we wanted higher decimal precision or higher amplitude of results we would need to trade one for the other, changing the TG_DECIMAL_BITS constant. No more changes would be needed.
sq: Same as tg, but you can change it separately. in case we want different compromises for each operation. 

The sin, cos, phase and modulus functions just use cordic and handle errors before they happen. 

The tan function uses sine and cosine and makes the division taking advantage of the tan = sin/cos property. 
It works on all four quadrants and handles division by 0 by returning the max value tg can have. (So technically the range of tg is a bit lower, as only the handle can return the max value).

The arctan uses the phase and draws the vector that tan defines.

### DIFF
--- a/Inc/HALAL/Services/CORDIC/CORDIC.hpp
+++ b/Inc/HALAL/Services/CORDIC/CORDIC.hpp
@@ -66,13 +66,13 @@ public:
 	 * If the angle or the out parameter has less size than the size parameter, it will throw an exception
 	 * If one or both angle and out parameters have more size than size, it will just calculate until size and leave the rest as it is
 	 */
-	static void cos(int32_t *angle,int32_t *out,int size);
+	static void cos(int32_t *angle,int32_t *out,int32_t size);
 
 
 	/*
 	 * @brief sine function. virtually the same as the cosine function
 	 */
-	static void sin(int32_t *angle,int32_t *out,int size);
+	static void sin(int32_t *angle,int32_t *out,int32_t size);
 
 
 	/*
@@ -89,7 +89,7 @@ public:
 	 * The exact time it takes is around 4 cycles more than the cos function, which is around 30-40. So 10% more as it needs to read one more register.
 	 * The reads are protected to give the correct result, thats why it needs 4 cycles for the read (on average) and not just one
 	 */
-	static void cos_and_sin(int32_t *angle, int32_t *cos_out, int32_t *sin_out, int size);
+	static void cos_and_sin(int32_t *angle, int32_t *cos_out, int32_t *sin_out, int32_t size);
 
 
 	/*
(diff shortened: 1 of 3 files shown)


StefanCostea:
[Math](https://trello.com/c/QUaEpFGT/65-math)

StefanCostea (approved):

Inc/HALAL/Services/CORDIC/CORDIC.hpp
```diff
@@ -66,13 +66,13 @@ class RotationComputer{
 	 * If the angle or the out parameter has less size than the size parameter, it will throw an exception
 	 * If one or both angle and out parameters have more size than size, it will just calculate until size and leave the rest as it is
 	 */
-	static void cos(int32_t *angle,int32_t *out,int size);
+	static void cos(int32_t *angle,int32_t *out,int32_t size);
 
 
 	/*
```
PabloGonzalezSegarra, line 72: Para doxygen hay que poner dos asteriscos al principio de cada comentario. 
```suggestion
	/**
```

Src/ST-LIB_LOW/Math/Math.cpp
```diff
@@ -0,0 +1,98 @@
...
+		if(((temp*temp)) <= over){
+			sol = sol + i;
+		}
+	}
+
+	return sol<<SQ_DECIMAL_BITS;
+}
+
+int32_t Math::sq_to_unitary(int32_t sq_in){
+	return sq_in << (32 - SQ_DECIMAL_BITS);
```
PabloGonzalezSegarra, line 84: Porque 32?

PabloGonzalezSegarra (approved):
Lo de la documentación hazlo en otra rama.

Src/ST-LIB_LOW/Math/Math.cpp
ricardochust: por que uso int de 32 bits. Son los que usa el cordic, así que no creo que vaya a cambiarlos. 
para obtener los bits que no son decimales resto a los bits totales los bits que son decimales, y luego shifteo todos esos fuera para dejar solo los decimales. Asi lo paso al rango [-1,1]. No hace falta decir que si le haces eso al numero 3 no va a guardar un 3, por eso de que no cabe en el tipo de unidad. Eso se queda solo con los decimales.

Inc/HALAL/Services/CORDIC/CORDIC.hpp
ricardochust: Ese cambio deberia hacerlo en el propio CORDIC, porque es de esa clase. Supongo que si lo hago aquí no debería haber problema ya que vamos a sacar el release así que todo se juntara, pero podriamos hacer una pull request para cosas del doxygen y documentacion.

PR #166, opened 2023-01-25
Added to_string method to Pin

### LINKED ISSUE #165: pin toString method
now that we have the ErrorHandler as our new tool in our belt we may need some ways to obtain information from our libraries so we can print them when the error happens.

The most obvious is the Pin, which we need to print so we can know where was the error. After a few tries and tricks it seems that the only way (or at least the most clean) to do this is by adding a toString method to Pin, which should give us at least the letter and the number. 
We may also be interested in getting the state of the pin, but we couls do that in another method an it is not as essential as getting the Pin itself

StefanCostea:
Added to_string method to Pin
This PR resolves #165


StefanCostea:
[Añadir to_String a Pin](https://trello.com/c/xIoCRHSK/93-a%C3%B1adir-tostring-a-pin)

ricardochust (approved):

Src/HALAL/Models/PinModel/Pin.cpp
```diff
@@ -136,6 +136,123 @@ PE4,PE5,PE6,PE7,PE8,PE9,PF0,PF1,PF10,PF11,PF12,PF13,PF14,PF15,PF2,PF3,PF4,PF5,PF
 PF8,PF9,PG0,PG1,PG10,PG11,PG12,PG13,PG14,PG15,PG2,PG3,PG4,PG5,PG6,PG7,PG8,PG9,PH0,PH1,
 PA2,PA3,PA4,PA5,PA6,PA7,PA8};
 
+map<Pin,string> Pin::map_to_string = {
```
PabloGonzalezSegarra, line 139: Hazlo constante.

PabloGonzalezSegarra (changes requested):

alejandrogonzalvo (approved):

Dgoncas (approved):

alejandrogonzalvo (approved):

PabloGonzalezSegarra (changes requested):
Primero el puerto.

PabloGonzalezSegarra (approved):

PR #182, opened 2023-01-31
Now only one parameter is necesary to call ErrorHandler

PabloGonzalezSegarra:
# Solution:
First we need to use `...` because `args...` needs atleast 1 parameter. Next we ned to use `##__VA_ARGS__` to pass the arguments list. The `##` is needed because in case the arguments list is empty will swallow the preceding comma.
```cpp
#define ErrorHandler(x, ...) do { ErrorHandlerModel::SetMetaData(__LINE__, __FUNCTION__, __FILE__); \
			          ErrorHandlerModel::ErrorHandlerTrigger(x, ##__VA_ARGS__);}while(0)

```
Now we need to change the method `ErrorHandlerModel::ErrorHandler(string format, Args.. args)` to `ErrorHandlerModel::ErrorHandlerTrigger(string format, ...)` to avoid redefinition due to de macro function and to accept an empty argument list.
```cpp
static void ErrorHandlerTrigger(string format, ...);
```

Inside the method we need to handle the variable list using `va_lists` and we need to change normal `snprintf` to `vsnprintf` because it use `va_list` as argument. We need to have in mind that we need to copy the `va_list arguments` if we want to use it twice. Simplified function: 
```cpp
void error(string format, ...){
    va_list arguments;
    va_start(arguments, format);
    va_list arg_copy;
    va_copy(arg_copy, arguments);

    const int32_t size = snprintf(nullptr, 0, format.c_str(), arguments) + 1;
    const unique_ptr<char[]> buffer = make_unique<char[]>(size);
    va_end(arguments);

    snprintf(buffer.get(), size, format.c_str(), arg_copy);
    va_end(arg_copy);

    description += string(buffer.get(), buffer.get() + size - 1);
}
```

### DIFF
--- a/Inc/C++Utilities/CppUtils.hpp
+++ b/Inc/C++Utilities/CppUtils.hpp
@@ -33,6 +33,8 @@
 #include <cstring>
 #include <span>
 #include <ranges>
+#include <cstdarg>
+#include <stdarg.h>
 
 namespace chrono = std::chrono;
 
--- a/Inc/ST-LIB_LOW/ErrorHandler/ErrorHandler.hpp
+++ b/Inc/ST-LIB_LOW/ErrorHandler/ErrorHandler.hpp
@@ -29,8 +29,7 @@ public:
 	 * @param args   Arguments specifying data to print
 	 * @return uint8_t Id of the service.
 	 */
-	template<class... Args>
-	static void ErrorHandler( string format, Args... args);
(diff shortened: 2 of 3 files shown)


StefanCostea (approved):

alejandrogonzalvo (approved):
Muy buen PR

ricardochust (approved):

PR #227, opened 2023-04-07
clean dma cache

alejandrogonzalvo:
This PR fixes Issue where DMA does not work with ETH

TODO: Testing

### DIFF
--- a/Src/HALAL/Services/ADC/ADC.cpp
+++ b/Src/HALAL/Services/ADC/ADC.cpp
@@ -15,6 +15,20 @@ extern ADC_HandleTypeDef hadc3;
 uint8_t ADC::id_counter = 0;
 unordered_map<uint8_t, ADC::Instance> ADC::active_instances = {};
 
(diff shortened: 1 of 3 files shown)


Src/HALAL/Services/ADC/ADC.cpp
```diff
@@ -15,6 +15,20 @@ extern ADC_HandleTypeDef hadc3;
 uint8_t ADC::id_counter = 0;
 unordered_map<uint8_t, ADC::Instance> ADC::active_instances = {};
 
+void HAL_ADC_ConvCpltCallback (ADC_HandleTypeDef *hadc){
```
StefanCostea, line 18: `	for(int i = 0; i<3; i++){
		if(hadc == ADC::peripherals[i].handle){
			SCB_CleanInvalidateDCache_by_Addr((uint32_t*)ADC::peripherals[i].dma_stream, 32);
			break;
		}
	}`

StefanCostea (approved):

PR #258, opened 2023-05-10
Fix/rework state machine

StefanCostea:
Added functionality in state machine to choose which kind of alarm to use in cyclic actions. 
Inverted some dependencies so it is easier to read and to expand for future types of alarms.


alejandrogonzalvo (approved):
Gran mejora de la state machine, me gusta mucho

ricardochust (approved):
se ve bien

PR #263, opened 2023-05-16
remove warnings

alejandrogonzalvo:
This PR removes all warnings from the ST-LIB


Inc/HALAL/Models/Packets/Packet.hpp
```diff
@@ -198,10 +198,11 @@ class HeapPacket : public Packet{
     }
 
     void parse(void* data) override {
-        data += sizeof(id);
+        byte* data_byte = (byte*)data;
```
PabloGonzalezSegarra, line 201: Porque no uint8_t?

PabloGonzalezSegarra (changes requested):
Hay que revisar la parte ethernet

ricardochust (changes requested):

PR #318, opened 2023-06-17
protection manager changes

ricardochust:
Changes added:

Changed IP of SNTP to the actual server IP (tested)

Put enum of faultType into a namespace so the global space isn t polluted with enums that are not meant to be used there (like FAULT)

Made that when the protection handler goes to fault it first sends a packet with ID 0 to all sockets to get all boards into fault faster. (tested)

add Error as a part of protection handler (tested)

changed build of notification to add a 0x00 byte at the end as the last version of chango-gui needs it to differenciate a packet end. (chango works with it and doesn t without in v12)

For now the ProtectionManager sends this fast packet once when it forces the GeneralStateMachine into fault. There is also a method called propagate_fault for anyone that needs it, but if any other behaviour is prefered for all boards let me know here.

image

### DIFF
--- a/Inc/ST-LIB_HIGH/Protections/Protection.hpp
+++ b/Inc/ST-LIB_HIGH/Protections/Protection.hpp
@@ -4,9 +4,9 @@
 #include "ErrorHandler/ErrorHandler.hpp"
 #include "Boundary.hpp"
 
-namespace protections{
+namespace Protections{
 enum FaultType{
-	FAULT,
+	FAULT = 0,
 	WARNING
 };
 }
@@ -17,7 +17,7 @@ private:
     char* name = nullptr;
     vector<unique_ptr<BoundaryInterface>> boundaries;
     BoundaryInterface* jumped_protection = nullptr;
-    static constexpr protections::FaultType fault_type = protections::FaultType::FAULT;
+    static constexpr Protections::FaultType fault_type = Protections::FaultType::FAULT;
 public:
     template<class Type, ProtectionType... Protector, template<class,ProtectionType> class Boundaries>
     Protection(Type* src, Boundaries<Type,Protector>... protectors) {
--- a/Inc/ST-LIB_HIGH/Protections/Protection.hpp
+++ b/Inc/ST-LIB_HIGH/Protections/Protection.hpp
@@ -4,10 +4,12 @@
 #include "ErrorHandler/ErrorHandler.hpp"
 #include "Boundary.hpp"
 
+namespace protections{
 enum FaultType{
 	FAULT,
 	WARNING
 };
+}
 
 class Protection{
 private:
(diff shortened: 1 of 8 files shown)


Src/ST-LIB_HIGH/Protections/ProtectionManager.cpp
```diff
@@ -5,6 +5,7 @@
 StateMachine* ProtectionManager::general_state_machine = nullptr;
 Notification ProtectionManager::fault_notification = {ProtectionManager::fault_id, ProtectionManager::to_fault};
```
StefanCostea, line 6: remove callback

Inc/ST-LIB_HIGH/Protections/Protection.hpp
```diff
@@ -4,18 +4,20 @@
 #include "ErrorHandler/ErrorHandler.hpp"
 #include "Boundary.hpp"
 
+namespace protections{
```
StefanCostea, line 7: use upper case

Src/ST-LIB_HIGH/Protections/ProtectionManager.cpp
```diff
@@ -5,6 +5,7 @@
...
 Notification ProtectionManager::warning_notification = {ProtectionManager::warning_id, nullptr};
+StackOrder<0> ProtectionManager::fault_order(0,to_fault);
```
StefanCostea, line 8: id should not be hardcoded

StefanCostea (changes requested):

Inc/ST-LIB_HIGH/ST-LIB_HIGH.hpp
```diff
@@ -23,3 +23,7 @@
 #include "Control/Blocks/MeanCalculator.hpp"
 #include "Control/ControlSystem.hpp"
 #include "FlashStorer/FlashStorer.hpp"
+
+namespace STLIB_HIGH {
```
StefanCostea, line 27: mete el stlib high start en el stlib start

StefanCostea (approved):

PabloGonzalezSegarra (approved):

Inc/ST-LIB_HIGH/ST-LIB_HIGH.hpp
ricardochust: > mete el stlib high start en el stlib start

hecho (esta en ST-LIB.cpp)

Src/ST-LIB_HIGH/Protections/ProtectionManager.cpp
ricardochust: done

ricardochust: Si quieres déjate esto hasta que haga los test mañana por la mañana y cuando confirme mergeamos
(o no, porque ya comprobé que funcionaba sin el de antemano, es una precaucion un poco excesiva pero a mi me mola asegurar)

PR #327, opened 2023-07-03
Fix/general

ricardochust:
Changed HAL start to use IPV4 instead of strings (didn t compile). ST-LIB start still works with strings and just cast them into IPV4 before sending it to HAL start

Solved SNTP parsing issue (needed a time distance lower than 36 years and RTC set too late for that, now changed)

Removed 0x00 at the end of notification, as it is no longer needed by the gui and messes with it

Added a minimun delay between protection notifications so they don t get send at the same speed as the check_protections run (can lead to failure if you try to send it to fast)

Changed Fault and Warning notification ids to properly work with the GUI

-------------------------------------------------------------------------

There is still a bug that forces into Hard Fault on line 300 of time.cpp (Time::high_precision_alarms_by_timer[tim].alarm();) and on line 48 of notification.hpp (if(buffer != nullptr) free(buffer);). The Hard Fault is a BusFault error (instruction prefetch or memory access failures) which are most probably caused by two consecutive free's or a malloc not having enough space

These don t occur on every run of a nucleo, instead occuring around 25% of the time you start the micro. Also, seems that the second one is solved by adding +100 to the get_string_size return value, which is very weird as the get_string_size method isn t even invocated to begin with. This bug will be attended on another pull request to not make this one too big. 

Also, trying to debug the Time::high_precision_alarms_by_timer[tim] results in a debugger crash.

If anyone sees any line of code on this pull request that could be at fault of any of those errors, put it on this pull request and do not accept it. 

https://interrupt.memfault.com/blog/cortex-m-hardfault-debug#bad-address-read-example
https://stackoverflow.com/questions/51057418/stm32-hardfault-at-free

### DIFF
--- a/Inc/HALAL/HALAL.hpp
+++ b/Inc/HALAL/HALAL.hpp
@@ -27,5 +27,5 @@
 #include "CORDIC/CORDIC.hpp"
 
 namespace HALAL {
-	void start(string ip, string subnet_mask, string gateway, UART::Peripheral& printf_peripheral);
+	void start(IPV4 ip, IPV4 subnet_mask, IPV4 gateway, UART::Peripheral& printf_peripheral);
 }
--- a/Inc/ST-LIB.hpp
+++ b/Inc/ST-LIB.hpp
@@ -6,6 +6,8 @@
 
 class STLIB {
 public:
+
+	static void start(IPV4 ip, IPV4 subnet_mask, IPV4 gateaway, UART::Peripheral& printf_peripheral);
 	static void start(
 			string ip = "192.168.1.4",
 			string subnet_mask = "255.255.0.0",
--- a/Inc/ST-LIB_HIGH/Protections/Notification.hpp
+++ b/Inc/ST-LIB_HIGH/Protections/Notification.hpp
@@ -12,7 +12,6 @@ class Notification : public Order{
 private:
 	typedef uint16_t message_size_t;
 	uint16_t id;
-	uint8_t end = 0x00;
     void(*callback)() = nullptr;
     string tx_message;
     message_size_t tx_message_size;
@@ -52,7 +51,6 @@ public:
     	memcpy(buffer,&id,sizeof(id));
     	memcpy(buffer + sizeof(id), &tx_message_size, sizeof(tx_message_size));
      	memcpy(buffer+sizeof(id)+sizeof(tx_message_size),tx_message.c_str(), tx_message.size());
-     	memcpy(buffer+sizeof(id)+sizeof(tx_message_size)+tx_message.size(),&end,sizeof(end));
     	return buffer;
     }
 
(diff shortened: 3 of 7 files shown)


ricardochust:
He estado haciendo multiples pruebas sobre el error de hard fault en el bus

He limpiado todo el espacio de trabajo para que haga una compilacion desde 0, y me daba el error del hard fault un 25% de las veces. 

Luego, para poder debuggear hice que me enviara el valor de contador tambien (la variable protegida), y por comodidad hice que lo enviara como el paquete de pressure de la TCU porque ya estaba en la gui. Para eso tuve que cambiar los paquetes de int a double; y cuando hice eso no volvio a aparecer el error mas. He hecho 20 runs seguidas y ha funcionado correctamente las 20. Lo unico que se me ocurre es que haya problemas con alocar los paquetes con variables de tipo int especificamente, pero la verdad es que no lo tengo muy claro. 

Sea como sea, con exactamente el mismo codigo en la ST-LIB y el template-project, pero cambiando las variables usadas en los paquetes de int a float o a double no hay problema. Hare pruebas para uint32_t y os digo.

ricardochust:
`#include "main.h"
#include "lwip.h"

#include "ST-LIB.hpp"
#include "Runes/Runes.hpp"

const IPV4 TCU_IP = {"192.168.0.14"};
const IPV4 BACKEND_IP = {"192.168.0.9"};

enum states{
	INITIAL = 0,
	OPERATIONAL,
	FAULT
};

double counter = 0;
uint8_t pstate = INITIAL;
uint8_t sstate = INITIAL;

void emergency_function(){
	counter++;
}

int main(void)
{
     	STLIB::start("192.168.0.14");

	StateMachine PrincipalStateMachine;
	PrincipalStateMachine.add_state(INITIAL);
	PrincipalStateMachine.add_state(OPERATIONAL);
	PrincipalStateMachine.add_state(FAULT);
	add_protection(&counter,Boundary<double,ABOVE>(10));
	ProtectionManager::link_state_machine(PrincipalStateMachine,FAULT);
	ProtectionManager::set_id(Boards::TCU);

	StackOrder emergency_state(1402,emergency_function);
	StackPacket state_machine_data(1403,&pstate,&sstate);
	StackPacket pressure_data(1404,&counter);

	DatagramSocket backUDP(TCU_IP,50400,BACKEND_IP,50400);
	ServerSocket backTCP(TCU_IP,50500);
	Time::register_high_precision_alarm(100000,ProtectionManager::check_protections);
	Time::register_low_precision_alarm(1,STLIB::update);
	Time::register_low_precision_alarm(100,[&](){pstate=PrincipalStateMachine.current_state;
	backUDP.send(state_machine_data);
	backUDP.send(pressure_data);
	});
	while(1) {

	}

}

void Error_Handler(void)
{
	ErrorHandler("HAL error handler triggered");
	while (1){}
}
`

PR #329, opened 2023-09-18
Fix/tcp queue

ricardochust:
added a method to queue packets so they get sent when the connection is accepted

added new method to tcp_client "reset", which hards reconnect into the server (destroys pcb and creates a new one)

yet to discuss behaviour when a FIN or RST close connection is received (should it retry reconnecting or throw an error into the Errorhandler)?

### DIFF
--- a/Inc/HALAL/Services/Communication/Ethernet/TCP/ServerSocket.hpp
+++ b/Inc/HALAL/Services/Communication/Ethernet/TCP/ServerSocket.hpp
@@ -15,6 +15,9 @@
 
 #define PBUF_POOL_MEMORY_DESC_POSITION 8
 
+/**
+* @brief ServerSocket class
+*/
 class ServerSocket : public OrderProtocol{
 public:
 
@@ -45,18 +48,32 @@ public:
 
 	void operator=(ServerSocket&& other);
 
+	/**
+	* @brief ends the connection between the server and the client. 
+	*/
 	void close();
 
+	/**
+	* @brief process the data received by the client orders. It is meant to be called only by Lwip on the receive_callback
+	*
+	* reads all the data received by the server in the ethernet buffer, packet by packet.
+	* Then, for each packet, it processes it depending on the order id (default behavior is not process) and removes it from the buffer. 
+	* This makes so the receive_callback (and thus the Socket) can only process declared orders, and ignores all other packets. 
+	*/
 	void process_data();
 
-	/*
-	 * @brief puts the order data into the tx_packet_buffer so it can be sent when a connection is accepted
+	/**
+	 * @brief saves the order data into the tx_packet_buffer so it can be sent when a connection is accepted
+	 *
+	 * @param order the order to send, which contains the data and id of the message
 	 * @return true if the data could be allocated in the buffer, false otherwise
 	 */
 	bool queue_order(Order& order);
 
-	/*
-	 * @brief puts the order data into the tx_packet_buffer and sends it
+	/**
+	 * @brief puts the order data into the tx_packet_buffer and sends all the data in the buffer to the client
+	 *
+	 * @param order the order to send, which contains the data and id of the message
 	 * @return true if the data was sent successfully, false otherwise
 	 */
 	bool send_order(Order& order) override{
@@ -85,8 +102,14 @@ public:
 		return true;
 	}
 
+	/**
+	* @brief sends all the binary data saved in the tx_packet_buffer to the connected client. 
+	*/
 	void send();
 
+	/**
+	* @return true if a connection with the client was established, false otherwise
+	*/
 	bool is_connected();
 
 private:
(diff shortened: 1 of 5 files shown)


Src/HALAL/Services/Communication/Ethernet/TCP/Socket.cpp
```diff
@@ -201,5 +243,23 @@ void Socket::error_callback(void *arg, err_t error){
...
+void Socket::connection_error_callback(void *arg, err_t error){
+	if(error == ERR_RST){
+		Socket* socket = (Socket*) arg;
+		socket->pending_connection_reset = true;
+		return;
+	}
+	ErrorHandler("Connection socket error: %d. Couldn t start client socket ", error);
+}
+
+err_t Socket::connection_poll_callback(void *arg, struct tcp_pcb* connection_control_block){
```
g0nz4I0, line 255: same as in **Socket::connection_error_callback**

Src/HALAL/Services/Communication/Ethernet/TCP/Socket.cpp
```diff
@@ -201,5 +243,23 @@ void Socket::error_callback(void *arg, err_t error){
 	ErrorHandler("Client socket error: %d. Socket closed",error);
 }
 
+void Socket::connection_error_callback(void *arg, err_t error){
```
g0nz4I0, line 246: maybe I am missing something but you pass a **void*** and instantly cast it to a **Socket*** by copying it.
also I assume it cannot be null, since you don't check for it. If that's the case just pass a **Socket&**, we avoid the copy and also avoid passing a **void***

Inc/HALAL/Services/Communication/Ethernet/TCP/Socket.hpp
```diff
@@ -83,5 +97,8 @@ class Socket : public OrderProtocol{
 	static err_t send_callback(void* arg, struct tcp_pcb* client_control_block, uint16_t length);
 	static void error_callback(void *arg, err_t error);
 
+	static err_t connection_poll_callback(void* arg, struct tcp_pcb* connection_control_block);
+	static void connection_error_callback(void *arg, err_t error);
```
g0nz4I0, line 101: why use void* as argument? please ensure type safety by passing a Socket*

Inc/HALAL/Services/Communication/Ethernet/TCP/Socket.hpp
```diff
@@ -83,5 +97,8 @@ class Socket : public OrderProtocol{
 	static err_t send_callback(void* arg, struct tcp_pcb* client_control_block, uint16_t length);
 	static void error_callback(void *arg, err_t error);
 
+	static err_t connection_poll_callback(void* arg, struct tcp_pcb* connection_control_block);
```
g0nz4I0, line 100: why use void* as argument? please ensure type safety by passing a Socket*

Inc/HALAL/Services/Communication/Ethernet/TCP/Socket.hpp
```diff
@@ -43,7 +46,18 @@ class Socket : public OrderProtocol{
 	void close();
 
 	void reconnect();
+	void reset();
 
+	/*
+	 * @brief puts the order data into the tx_packet_buffer so it can be sent when a connection is accepted
+	 * @return true if the data could be allocated in the buffer, false otherwise
+	 */
+	bool queue_order(Order& order);
```
g0nz4I0, line 55: same as in ServerSocket.hpp queue_order

Inc/HALAL/Services/Communication/Ethernet/TCP/ServerSocket.hpp
```diff
@@ -45,10 +48,34 @@ class ServerSocket : public OrderProtocol{
...
+	*/
 	void process_data();
 
+	/**
+	 * @brief saves the order data into the tx_packet_buffer so it can be sent when a connection is accepted
+	 *
+	 * @param order the order to send, which contains the data and id of the message
+	 * @return true if the data could be allocated in the buffer, false otherwise
+	 */
+	bool queue_order(Order& order);
```
g0nz4I0, line 71: I don't like the naming of this, I understand where it is coming from, "queuing" as a verb, but I believe something like:
_add_order_to_queue()_ is more clear and easier to grasp

Src/HALAL/Services/Communication/Ethernet/TCP/Socket.cpp
ricardochust: the lwip library needs to call this function in order for it to work, and the way its made is that the function has to be passed with a setter function (in this case tcp_err) after you construct the pcb. 

The function the tcp_err receives as a parameter has to be specifically a null pointer + an err_t value, in order for it to work. Passing any other type of function would lead into an invalid conversion error.

Inc/HALAL/Services/Communication/Ethernet/TCP/Socket.hpp
ricardochust: done

Src/HALAL/Services/Communication/Ethernet/TCP/Socket.cpp
ricardochust: i closed the other change request so we can talk it in this one. If you have an idea to get around the lwip type problem put it here. If you are fine with the explanation you are free to close this one too.

g0nz4I0 (approved):
LGTM!

AmancioG (approved):

PR #332, opened 2023-10-09
Feature/cmake build

g0nz4I0:
Added support for building with cmake

Ideally place the stlib under /opt/STLIB and give permissions to the user

```
sudo cp -r STLIB /opt
sudo chmod a+rwx /opt/STLIB
```
This way we can easily edit the STLIB and the build script can move around files

Added a build script in tools/ this script supports:

- selecting the build configuration `-bb [Debug, Release]`
- choosing the target `-t [BOARD, NUCLEO]`
- enabling Ethernet via ifdef guard `-eth [ON,OFF]`

image

Minor changes: 

- made CMakeLists.txt visible in project explorer
- the output folder is now build/<_CONFIGURATION_>/lib

this branch also implements the minor changes to the time class


CMakeLists.txt
```diff
@@ -53,7 +70,7 @@ target_compile_options(${LIBRARY} PUBLIC
   $<$<COMPILE_LANGUAGE:CXX>:-fno-rtti>
 
   -Wno-psabi
-  --specs=nosys.specs
+  -specs=/usr/lib/arm-none-eabi/lib/nosys.specs
```
alejandrogonzalvo, line 73: Does this work with a single hyphen? Be sure to test it with real programs that blink leds and things like that

CMakeLists.txt
```diff
@@ -1,16 +1,44 @@
...
 project (ST-LIB)
 
 set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED True)
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
-set(LIBRARY st-lib)
 
+set(LIBRARY st-lib)
+  
 file(GLOB_RECURSE SOURCE_C ${CMAKE_SOURCE_DIR}/ *.c)
```
alejandrogonzalvo, line 13: This is wrong. As I mentioned you in person, there is the space here, remove it for now so it actually matches the correct regex. I recommend you to remove the regex and manually input all the files. Or even better as you moved it to python you could dinamically find the source files using a regex in python and modify this file from there. The difference in using a python regex is that you can create you own feedback output, while GLOB does not output the regex.

alejandrogonzalvo (changes requested):
See comment on GLOB regex, this does not correctly compile ST-LIB

alejandrogonzalvo:
Also, last year lacked a lot of manual testing. Have a bunch of test programs, put them in the board and check everything is alright. It's not the HIL, its not fun, but it works. Id recommend you using this initial months to do this functional testing and also compare binary sizes between Cube and CMake until the difference is negligble. If the tool is not reliable by the end of the design phase in December, developers will be reluctant to use it and will end up just using Cube whenever they have the possibility.

CMakeLists.txt
g0nz4I0: I have tested it with a nucleo and seems to work fine, also from the gcc page: 
image

CMakeLists.txt
g0nz4I0: I understand what you are saying but look at this, 
on the right there is the build removing the whitespaces, on the left the build with whitespaces, it does not compile any more files.
image

 Do you think if there were extra .cpp files it would compile them?
It does not search for all of the cpp files, NOTE that this happens with the following line: `file(GLOB_RECURSE SOURCE_CPP ${CMAKE_SOURCE_DIR}/ /*.cpp)` with the second slash it is clear that it searches starting from / but in the case of the whitespace seems to ignore it

However I agree that the GLOB approach is not ideal and I will manually add the files.

alejandrogonzalvo: Very interesting, did not know :+1:

CMakeLists.txt
```diff
@@ -22,28 +50,17 @@ target_compile_definitions(${LIBRARY} PUBLIC
...
-endif()
 
 
 target_compile_options(${LIBRARY} PUBLIC
   -mcpu=cortex-m7
   -mfpu=fpv5-d16
   -mfloat-abi=hard
   -mthumb
-  
   -Og
```
alejandrogonzalvo, line 64: If this is hardcoded, there is no use in setting the option above. Change also -g3
```suggestion
  -${OPTIMIZATION}
```

alejandrogonzalvo (changes requested):

alejandrogonzalvo (approved):
Nice work!! 
Still don't get why we don't like the Rust program :(. Let's hope this Python language is reliable enough :monocle_face:

CMakeLists.txt
g0nz4I0: fixed

g0nz4I0:
> Nice work!! Still don't get why we don't like the Rust program :(. Let's hope this Python language is reliable enough 🧐

the python script works even on ArchLinux 🎉, might also be able to make it work in Windows 
jokes aside, this way the build script is relative to the project, both the path and also the contents themself
However the credit is yours, I just moved a rust script into a python one

AmancioG (approved):

PR #339, opened 2023-11-11
Spi slave with DMA

ricardochust:
**There are changes on the template-project side, in the spi dma branch. The code on this branch won t compile unless the changes on the spi dma branch are on, so after merging this branch we will need to merge the branch on the template-project too.** 

This pull request changes how SPI behaves to use DMA and Interrupts described inside the class. On the user interface, he starts a receive or transmit communication; on the former the data is saved in the given span, while the latter uses DMA to take the memory and send it, triggering the interrupt on end of communication on both cases.

To use SPI slave or SPI master simply change the configuration of the peripheral mode on runes, or modify it by code before inscribing it. 

+Added SPIPackets and master_send_packet function (which sends a single packet) and slave_receive_packets function (which starts running the slave in the background and updates the packets values to the last packet of that type received)

Testing code:
```
//#define MASTER

uint8_t* data;
SPI::Instance* inst;

int main(void)
{
	uint8_t spi_server = SPI::inscribe(SPI::spi3);
	inst = SPI::registered_spi[spi_server];

	SPIPacket hello(1000,2,2);
	data = &hello.master_data[0];
	STLIB::start();

#ifdef MASTER
	Time::register_low_precision_alarm(1, [&](){
			SPI::master_transmit_packet(spi_server, hello
					);});
#else
	SPI::slave_listen_packets(spi_server);
#endif

	while(1){
#ifdef MASTER
		hello.master_data[3] = 15;
#else
		hello.slave_data[3] = hello.master_data[3] + 1;
#endif
		STLIB::update();
	}
}
```

### DIFF
--- a/Inc/HALAL/Models/Packets/SPIPacket.hpp
+++ b/Inc/HALAL/Models/Packets/SPIPacket.hpp
@@ -1,0 +1,54 @@
+/*
+ * SpiPacket.hpp
+ *
+ *  Created on: 15 nov. 2023
+ *      Author: ricardo
+ */
+#pragma once
+
+#include "ErrorHandler/ErrorHandler.hpp"
+
+
+class SPIPacket{
+public:
+	static map<uint16_t, SPIPacket*> SPIPacketsByID;
+	uint16_t id;
+	uint8_t* master_data;
+	uint16_t master_data_size;
+	uint8_t* slave_data;
+	uint16_t slave_data_size;
+	uint16_t greater_data_size;
+
+	/**
+	 * @brief constructor of SPIPacket. The SPIPacket uses the first byte of the data arrays to coordinate the packet id, and builds the arrays with 1 more byte of space.
+	 *
+	 */
+	SPIPacket(uint16_t id, uint16_t master_data_size, uint16_t slave_data_size) : id(id), master_data_size(master_data_size), slave_data_size(slave_data_size){
+		if(id == 0){
+			ErrorHandler("Used reserved ID 0 on SPIPacket");
+		}
+		if(master_data_size > slave_data_size){
+			greater_data_size = master_data_size+2;
+		}else{
+			greater_data_size = slave_data_size+2;
+		}
+		master_data = new uint8_t[greater_data_size];
+		slave_data = new uint8_t[greater_data_size];
+		master_data[0] = (uint8_t) id;
+		master_data[1] = (uint8_t) (id>>8);
+		slave_data[0] = (uint8_t) id;
+		slave_data[1] = (uint8_t) (id>>8);
+		SPIPacket::SPIPacketsByID[id] = this;
+	}
+
+	SPIPacket(uint16_t id, uint8_t* master_data, uint8_t* slave_data, uint16_t shared_data_size) :
+		id(id), master_data(master_data), master_data_size(shared_data_size), slave_data(slave_data), slave_data_size(shared_data_size){
+		if(id == 0){
+			ErrorHandler("Used reserved ID 0 on SPIPacket");
+		}
+
+	}
+
+};
+
+
--- a/Inc/HALAL/Models/Packets/SPIPacket.hpp
+++ b/Inc/HALAL/Models/Packets/SPIPacket.hpp
@@ -25,7 +25,7 @@ public:
 	 */
 	SPIPacket(uint16_t id, uint16_t master_data_size, uint16_t slave_data_size) : id(id), master_data_size(master_data_size), slave_data_size(slave_data_size){
 		if(id == 0){
-			ErrorHandler("Used reserved ID 0 on SPIPacket");
+			ErrorHandler("Cannot use 0 as the SPIPacketID, as it is reserved to the no packet ready signal");
 		}
 		if(master_data_size > slave_data_size){
 			greater_data_size = master_data_size+2;
@@ -44,7 +44,7 @@ public:
 	SPIPacket(uint16_t id, uint8_t* master_data, uint8_t* slave_data, uint16_t shared_data_size) :
 		id(id), master_data(master_data), master_data_size(shared_data_size), slave_data(slave_data), slave_data_size(shared_data_size){
 		if(id == 0){
-			ErrorHandler("Used reserved ID 0 on SPIPacket");
+			ErrorHandler("Cannot use 0 as the SPIPacketID, as it is reserved to the no packet ready signal");
 		}
 
 	}
--- a/Inc/HALAL/Models/Packets/SPIPacket.hpp
+++ b/Inc/HALAL/Models/Packets/SPIPacket.hpp
@@ -4,6 +4,7 @@
  *  Created on: 15 nov. 2023
  *      Author: ricardo
  */
+
 #pragma once
 
 #include "ErrorHandler/ErrorHandler.hpp"
--- a/Inc/HALAL/Services/Communication/SPI/SPI.hpp
+++ b/Inc/HALAL/Services/Communication/SPI/SPI.hpp
@@ -176,10 +176,19 @@ public:
 	 */
     static bool transmit_and_receive(uint8_t id, span<uint8_t> command_data, span<uint8_t> receive_data);
 
+    /**
+     * @brief update that has to be called in order for master to check if the slave is ready to send the packet. If it is not called periodically, the master_transmit_packet will not work. Not needed for dummy communication (not using packets)
+     */
     static void packet_update();
 
+    /**
+     * @brief master send packet method, which tries to send a single packet
+     */
     static bool master_transmit_packet(uint8_t id, SPIPacket packet);
 
+    /**
+     * @brief slave listen packets method. When called, the slave will start to listen packets until the state is set again to IDLE
+     */
     static void slave_listen_packets(uint8_t id);
 
     /**
--- a/Inc/HALAL/Services/Communication/SPI/SPI.hpp
+++ b/Inc/HALAL/Services/Communication/SPI/SPI.hpp
@@ -10,6 +10,7 @@
 #include "PinModel/Pin.hpp"
 #include "Packets/RawPacket.hpp"
 #include "DigitalOutputService/DigitalOutputService.hpp"
+#include "DMA/DMA.hpp"
 #include "ErrorHandler/ErrorHandler.hpp"
 
 #ifdef HAL_SPI_MODULE_ENABLED
@@ -36,6 +37,8 @@ private:
 
         SPI_HandleTypeDef* hspi;  /**< HAL spi struct pin. */  
         SPI_TypeDef* instance; /**< HAL spi instance. */
+        DMA::Stream hdma_tx; /**< HAL DMA handler for writting */
+        DMA::Stream hdma_rx; /**< HAL DMA handler for reading */
 
         uint32_t baud_rate_prescaler; /**< SPI baud prescaler.*/
         uint32_t mode = SPI_MODE_MASTER; /**< SPI mode. */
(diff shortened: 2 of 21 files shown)


Inc/HALAL/Models/Packets/SPIPacket.hpp
```diff
@@ -0,0 +1,54 @@
...
+	uint16_t slave_data_size;
+	uint16_t greater_data_size;
+
+	/**
+	 * @brief constructor of SPIPacket. The SPIPacket uses the first byte of the data arrays to coordinate the packet id, and builds the arrays with 1 more byte of space.
+	 *
+	 */
+	SPIPacket(uint16_t id, uint16_t master_data_size, uint16_t slave_data_size) : id(id), master_data_size(master_data_size), slave_data_size(slave_data_size){
+		if(id == 0){
+			ErrorHandler("Used reserved ID 0 on SPIPacket");
```
g0nz4I0, line 28: clarify ErrorHandler message, smth like "0 is a reserved ID "

Inc/HALAL/Services/Communication/SPI/SPI.hpp
```diff
@@ -8,24 +8,37 @@
...
+		PROCESSING_PACKET,
+    };
+
     /**
-     * @brief Struct which defines all data refering to SPI peripherals. It is
-     *        declared private in order to prevent unwanted use. Only 
-     *        predefined instances should be used.
+     * @brief Struct which defines all data referring to SPI peripherals. DO NOT MODIFY ON RUN TIME.
+     * declared public so HAL callbacks can directly access the instance
+     * NOT intended to be modified after the inscribe of the spi, configuration should be made in Runes.hpp
```
g0nz4I0, line 41: maybe  consider declaring it as private? so it is properly encapsulated?

Src/HALAL/HALAL.cpp
```diff
@@ -58,7 +58,7 @@ void HALAL::start(IPV4 ip, IPV4 subnet_mask, IPV4 gateway, UART::Peripheral& pri
 #endif
 
 #ifdef HAL_ETH_MODULE_ENABLED
-	Ethernet::start(ip, subnet_mask, gateway);
+	//Ethernet::start(ip, subnet_mask, gateway);
```
g0nz4I0, line 61: 😞 gud ol `git add .`

Src/HALAL/Models/Packets/SPIPacket.cpp
```diff
@@ -0,0 +1,10 @@
+/*
+ * SpiPacket.hpp
+ *
+ *  Created on: 15 nov. 2023
+ *      Author: ricardo
+ */
+
+#include "HALAL/Models/Packets/SPIPacket.hpp"
+
+map<uint16_t, SPIPacket*> SPIPacket::SPIPacketsByID = {};
```
g0nz4I0, line 10: Pro tip: `map<uint16_t, SPIPacket*> SPIPacket::SPIPacketsByID {};`

should work the same

Src/HALAL/Services/Communication/SPI/SPI.cpp
```diff
@@ -4,12 +4,14 @@
...
  *      Author: Pablo
  */
+
 #include "Communication/SPI/SPI.hpp"
 
 #ifdef HAL_SPI_MODULE_ENABLED
 
 
-unordered_map<uint8_t, SPI::Instance*> SPI::registered_spi = {};
+map<uint8_t, SPI::Instance*> SPI::registered_spi = {};
```
g0nz4I0, line 13: same as before

Src/HALAL/Services/Communication/SPI/SPI.cpp
```diff
@@ -47,42 +54,49 @@ bool SPI::transmit(uint8_t id, uint8_t data){
...
+		 	 return true;
+		 break;
+	 	 case HAL_BUSY:
+	 		 return false;
+	 	 break;
+	 	 default:
+	 		 ErrorHandler("Error while transmiting and receiving with spi DMA. Errorcode %u",(uint8_t)errorcode);
+	 		 return false;
+	 	 break;
+	 }
```
g0nz4I0, line 76: no need to break after returning, break is normally used in switch statements because by default swith statements fallthrough

Src/HALAL/Services/Communication/SPI/SPI.cpp
```diff
@@ -47,42 +54,49 @@ bool SPI::transmit(uint8_t id, uint8_t data){
...
```
g0nz4I0, line 99: no need to break after returning, break is normally used in switch statements because by default swith statements fallthrough

Src/ST-LIB.cpp
```diff
@@ -21,6 +21,7 @@ void STLIB::start(string ip, string subnet_mask, string gateaway,  UART::Periphe
 }
 
 void STLIB::update() {
-    Ethernet::update();
+    //Ethernet::update();
```
g0nz4I0, line 24: 😞

Src/ST-LIB.cpp
```diff
@@ -21,6 +21,7 @@ void STLIB::start(string ip, string subnet_mask, string gateaway,  UART::Periphe
...
 	ErrorHandlerModel::ErrorHandlerUpdate();
+	SPI::packet_update();
```
g0nz4I0, line 26: I believe this should be left up to the user to place it there, I understand that by default the map is empty, yet it bloats the function when in reality it is doing nothing

g0nz4I0:
Also, create the PR with the changes that you mention in the template project, otherwise no one can use it :)

Inc/HALAL/Models/Packets/SPIPacket.hpp
ricardochust: 👍

Inc/HALAL/Services/Communication/SPI/SPI.hpp
ricardochust: HAL callbacks need to access the instance for multiple operations, which are not part of the class (they are overriden functions of the HAL), and they are run in an interruption two times for each packet, so i would prefer to add the less layers possible there. 

Also a user could be interested in modifying it _before_ the inscribe on the spi (which is not against its use) so he can have only one runes for two different boards, which change only the configuration from slave to master by using a flag.

Src/HALAL/HALAL.cpp
```diff
@@ -10,7 +10,7 @@
 void HALAL::start(IPV4 ip, IPV4 subnet_mask, IPV4 gateway, UART::Peripheral& printf_peripheral) {
 
 #ifdef HAL_ETH_MODULE_ENABLED
-	Ethernet::inscribe();
+	//Ethernet::inscribe();
```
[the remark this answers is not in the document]
ricardochust: solved (it was a conflict on branches)

Src/HALAL/HALAL.cpp
ricardochust: solved (it was a conflict on branches)

Src/HALAL/Services/Communication/SPI/SPI.cpp
ricardochust: first comment conversation applies to all the break comments

Src/HALAL/Services/Communication/SPI/SPI.cpp
```diff
@@ -91,25 +105,68 @@ bool SPI::transmit_and_receive(uint8_t id, span<uint8_t> command_data, span<uint
...
+
+void SPI::packet_update(){
+	for(auto iter: SPI::registered_spi){
+		if(iter.second->SPIPacketID != 0 && iter.second->available_end != iter.second->SPIPacketID && iter.second->state == STARTING_PACKET && iter.second->mode == SPI_MODE_MASTER && Time::get_global_tick() - iter.second->last_end_check > MASTER_SPI_CHECK_DELAY){
+			turn_off_chip_select(iter.second);
+			HAL_SPI_TransmitReceive_DMA(iter.second->hspi, (uint8_t *)&iter.second->SPIPacketID, (uint8_t *)&iter.second->available_end, 2);
+			iter.second->last_end_check = Time::get_global_tick();
+		}
 	}
+}
```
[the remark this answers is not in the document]
ricardochust: this one with the SPI::Packet_update() are work in progress (as i will need to test again after making the changes). 
The rest are marked as solved by the reasons put on each comment. 

As an extra point, structure bindings look nice 🔥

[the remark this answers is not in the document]
g0nz4I0: Perfect, the comments are mostly things to keep in mind, they are not critial, good job

Src/HALAL/Services/Communication/SPI/SPI.cpp
```diff
@@ -91,25 +105,68 @@ bool SPI::transmit_and_receive(uint8_t id, span<uint8_t> command_data, span<uint
...
+	spi->state = SPI::STARTING_PACKET;
+	spi->SPIPacketID = packet.id;
+	SPI::chip_select_off(id);
+	HAL_SPI_TransmitReceive_DMA(spi->hspi, (uint8_t *)&spi->SPIPacketID, (uint8_t *)&spi->available_end, 2);
+	return true;
+}
+
+void SPI::packet_update(){
+	for(auto iter: SPI::registered_spi){
+		if(iter.second->SPIPacketID != 0 && iter.second->available_end != iter.second->SPIPacketID && iter.second->state == STARTING_PACKET && iter.second->mode == SPI_MODE_MASTER && Time::get_global_tick() - iter.second->last_end_check > MASTER_SPI_CHECK_DELAY){
```
StefanCostea, line 147: add a comment explaining this line

Src/HALAL/Services/Communication/SPI/SPI.cpp
```diff
@@ -24,9 +24,14 @@ uint8_t SPI::inscribe(SPI::Peripheral& spi){
 
 	SPI::Instance* spi_instance = SPI::available_spi[spi];
 
+	spi_instance->SCK->alternative_function = AF5;
+	spi_instance->MOSI->alternative_function = AF5;
+	spi_instance->MISO->alternative_function = AF5;
```
hyperloop-orangeman, line 29: what is this AF5?

Src/HALAL/HALAL.cpp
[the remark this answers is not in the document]
hyperloop-orangeman: xd, it has come back to haunt, dont worry, I fix it

hyperloop-orangeman (approved):

g0nz4I0 (approved):

Src/HALAL/Services/Communication/SPI/SPI.cpp
ricardochust: it doesn t do anything yet, as the alternative function pin declaration for spi is hard coded in the stm32_h7xx_hal_msp.c and is not handled by the ST-LIB (most of the modules of the ST-LIB do this (for now)), which makes the alternate function argument of the Pin unused. 
_(in reality the HAL methods on the stm32_h7xx_hal_msp.c are called by the ST-LIB, they just don t use the Pin structure inside them as you cannot pass them such parameters)_
This is a placeholder for a future update where we stop using the HAL initialization functions (if we decide to do that); and a way to synchronize the structure configuration to the actual hardcoded hardware configuration used by the peripheral. These lines being there or not should not change the behavior of the module (i tested it and it works on my computer), so in reality they could be removed, but i find weird having the Pin structure display AF0 when in reality the Pin is running AF5.

PR #340, opened 2023-11-15
removed str constructor as IPV4 can resolve

ricardochust:
When trying to make a ServerSocket using a string C++ couldn t resolve as both IPV4 and string constructors can accept an string. 

Just removed the string constructor, which just called the IPV4 one. Tested with the wiki ServerSocket code (changed ip to 192.168.0.3)


AmancioG (approved):

g0nz4I0 (approved):
Will accept, but in case you really wanted to use a string as the IP I guess you could use:
auto* ss = new ServerSocket(string{"192.168.x.x"},6969);
because that would mean no narrowing, so the compiler will pick the string constructor
image
image
Whats happening is that both string and IPV4 have a constructor that takes in const char*, so it cannot resolve, will leave up to you if you want to merge

StefanCostea (approved):

PR #341, opened 2023-11-18
enhancement to DigitalIn sensors

g0nz4I0:
added` int get_value() `function, which reads and updates the value contained, and then returns the value, this allows for better encapsulation, not having to access directly to the variable


g0nz4I0:
ignore branch name :)

PR #345, opened 2023-12-01
feature/fix build script

g0nz4I0:
Fixed CMake error on post build on Windows and removed absolute path for nosys.specs file, also Windows build is now supported

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,7 +72,7 @@ target_compile_options(${LIBRARY} PUBLIC
   $<$<COMPILE_LANGUAGE:CXX>:-fno-rtti>
 
   -Wno-psabi
-  -specs=/usr/lib/arm-none-eabi/lib/nosys.specs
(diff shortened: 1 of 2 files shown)


AmancioG:
You can use:
 
       import os
       path = os.path.join(arg1, arg2, arg3, ... )

and writting the path in the arguments, you will get the OS default direcotry separator automatically. For example:

path = os.path.join("folder1", "folder2", "forlder3", "file")

path, on posix systems, will be:
       "folder/folder2/folder3/file"

and on Windows, it will instead be:
        "folder\folder2\folder3\file"

g0nz4I0:
Addressed, this would have made it much easier from the beginning: I don't know why I didn't use it :(

g0nz4I0:
Test slack bot

ricardochust (approved):

ricardochust:
it works on my machine 👍

AmancioG (approved):

Issue #346, opened 2023-12-03
Review DigitalIn Sensor


g0nz4I0:
While testing with the TUD I encountered weird behaviours with the current read method, so I went ahead and added a small hotfix, please check if this issue exists, if so I have already implemented the fix

PR #347, opened 2023-12-07
Fix/digital sensor

g0nz4I0:
This is regarding #346 take a look

### DIFF
--- a/Inc/ST-LIB_LOW/Sensors/DigitalSensor/DigitalSensor.hpp
+++ b/Inc/ST-LIB_LOW/Sensors/DigitalSensor/DigitalSensor.hpp
@@ -15,14 +15,15 @@
 class DigitalSensor{
 public:
 	DigitalSensor() = default;
-	DigitalSensor(Pin &pin, PinState *value);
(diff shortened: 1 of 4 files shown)


Inc/ST-LIB_LOW/Sensors/DigitalSensor/DigitalSensor.hpp
```diff
@@ -15,14 +15,15 @@
 class DigitalSensor{
 public:
 	DigitalSensor() = default;
-	DigitalSensor(Pin &pin, PinState *value);
-	DigitalSensor(Pin &pin, PinState &value);
+	DigitalSensor(Pin &pin, PinState value);
+	//DigitalSensor(Pin &pin, PinState &value);
```
StefanCostea, line 19: Left commented?

StefanCostea (changes requested):

g0nz4I0:
Won't merge, previous DigitalSensor works fine, and the approach of taking a variable to hold the value might be easier to use with protections than encapsulating the value in the sensor

PR #354, opened 2023-12-24
Development

g0nz4I0:
The purpose of this PR is to have a cleaner gitflow, and to not leave `main` behind, quickly review the changes as they are only from develop.

Reminder: 
open issues whenever you feel something is missing/ needs to be fixed.

🚀 🚀


ricardochust (approved):
we need to check what happens with sntp (i assume it is commented for a reason)
I will open an issue for it and accept this. 

Also make sure that the template_project has the changes for spi_dma when this is pushed

g0nz4I0:
> we need to check what happens with sntp (i assume it is commented for a reason) I will open an issue for it and accept this.
> 
> Also make sure that the template_project has the changes for spi_dma when this is pushed

SPI_DMA changes are already in template project if I recall correctly, about SNTP let me review it and maybe we can do some tests? Honestly I did not see it was commented out

AmancioG (approved):

PR #357, opened 2023-12-28
Protections refactor

g0nz4I0:
Bunch of changes review carefully:

RTC:
-   Created a global static entity  to allow direct access and usage as a variable within a Packet
-   Moved to a separate file (minor stuff)

Boundary:
 - Sending now uses standard packets, containing _(roughly)_: 
```
    id(TBD) | format_id | boundary_type_id | name_len | name |
    < Boundary metrics> |     < Boundary metrics> |     < Boundary metrics> |
   RTC data |    RTC data |    RTC data |    RTC data |    RTC data |
   RTC data |    RTC data |    RTC data |    RTC data |    RTC data |
```
Protection Manager:
- Added an `initialize()` method that sets the name of all the boundaries registered, since it does not make any sense to have to do it every time you want to send it

Discarded the notification class, still have to revisit the functionality,
Unified the TIME_ACCUMULATION specialization instead of having the same one for doubles and floats

### DIFF
--- a/Inc/ST-LIB_HIGH/Protections/Boundary.hpp
+++ b/Inc/ST-LIB_HIGH/Protections/Boundary.hpp
@@ -2,9 +2,15 @@
 #define PROTECTIONTYPE_LENGTH 7
 #include "ErrorHandler/ErrorHandler.hpp"
 #include "Control/Blocks/MeanCalculator.hpp"
+#include "HALAL/Models/Packets/Order.hpp"
+#include "HALAL/Services/Time/RTC.hpp"
 
-enum ProtectionType : uint64_t {
-    BELOW = std::numeric_limits<uint64_t>::max() - PROTECTIONTYPE_LENGTH + 1,
+using type_id_t = void(*)();
+template<typename>
+void type_id() {}
+
+enum ProtectionType : uint8_t {
+    BELOW = 0,
     ABOVE,
     OUT_OF_RANGE,
     EQUALS,
@@ -16,16 +22,22 @@ enum ProtectionType : uint64_t {
 struct BoundaryInterface{
 public:
     virtual bool check_bounds() = 0;
-	virtual char* serialize(char* dst) = 0;
-	virtual int get_string_size() = 0;
+	HeapOrder* message;
 
+	void update_name(char* n){
+		name = n;
+	}
 protected:
+	static const map<type_id_t,uint8_t> format_look_up;
 	static int get_error_handler_string_size(){
 		return ErrorHandlerModel::description.size();
 	}
 	static string get_error_handler_string(){
 		return ErrorHandlerModel::description;
 	}
+	//this will store the name of the variable
+	char* name;
+	static constexpr size_t NAME_MAX_LEN = 30;
 };
 
 template<class Type, ProtectionType Protector> struct Boundary;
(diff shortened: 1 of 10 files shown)


Src/ST-LIB_HIGH/Protections/Boundary.cpp
```diff
@@ -0,0 +1,16 @@
+#include "ST-LIB_HIGH/Protections/Boundary.hpp"
+
+ const map<type_id_t,uint8_t> BoundaryInterface::format_look_up{
+    {type_id<int>,0},
```
StefanCostea, line 4: Maybe using an Enum here for clariry

Src/ST-LIB_HIGH/Protections/ProtectionManager.cpp
```diff
@@ -81,29 +81,27 @@ void ProtectionManager::check_high_frequency_protections(){
...
+        ProtectionManager::notify(protection);
     }
 }
 
 void ProtectionManager::warn(string message){
 	warning_notification.notify(message);
 }
 
+void ProtectionManager::notify(Protection& protection){
+    if(protection.jumped_protection->Protector == ERROR_HANDLER){
```
StefanCostea, line 97: un poco ñapa xd pero lo compro

Inc/ST-LIB_HIGH/Protections/Boundary.hpp
```diff
@@ -16,246 +22,224 @@ enum ProtectionType : uint64_t {
...
-	virtual int get_string_size() = 0;
+	HeapOrder* message;
 
+	void update_name(char* n){
+		name = n;
+		string_len = name.size();
+	}
+	virtual void update_error_handler_message( [[maybe_unused]] const char* err_message){
+	}
+	ProtectionType Protector = ERROR_HANDLER;
```
StefanCostea, line 33: ??

Inc/ST-LIB_HIGH/Protections/Boundary.hpp
```diff
@@ -16,246 +22,224 @@ enum ProtectionType : uint64_t {
 struct BoundaryInterface{
 public:
     virtual bool check_bounds() = 0;
-	virtual char* serialize(char* dst) = 0;
-	virtual int get_string_size() = 0;
+	HeapOrder* message;
 
+	void update_name(char* n){
```
StefanCostea, line 27: check if max_len is surpassed

Inc/ST-LIB_HIGH/Protections/Boundary.hpp
```diff
@@ -16,246 +22,224 @@ enum ProtectionType : uint64_t {
...
+	{
+		
+	}
+	Boundary(Type* src, Boundary<Type, Protector> boundary): 
+		src(src),boundary(boundary.boundary)
+		{
+			format_id = BoundaryInterface::format_look_up.at(type_id<Type>);
+			// we have to preallocate space, otherwise the might get moved around, invalidating the pointer, better safe than sorry
+			name.reserve(NAME_MAX_LEN);
+			message = new HeapOrder(uint16_t{111},&format_id,&boundary_type_id,&string_len,&name,&this->boundary,this->src,
```
StefanCostea, line 71: magic 111

Inc/ST-LIB_HIGH/Protections/Boundary.hpp
```diff
@@ -16,246 +22,224 @@ enum ProtectionType : uint64_t {
...
-		return dst;
+	void update_error_handler_message(const char* err_message)override{
+		error_handler_string = err_message;
+		error_handler_string_len = error_handler_string.size();
 	}
+	private:
+		uint8_t boundary_type_id = Protector;
+		string error_handler_string;
+		uint16_t error_handler_string_len;
+		static constexpr uint16_t ERROR_HANDLER_MSG_MAX_LEN = 255;
```
StefanCostea, line 203: why would you set a limit?

StefanCostea: TCP Ethernet packets can be at most 1458 bytes if you want to avoid segementain, will still work though

Inc/ST-LIB_HIGH/Protections/Boundary.hpp
```diff
@@ -16,246 +22,224 @@ enum ProtectionType : uint64_t {
...
-struct Boundary<float, TIME_ACCUMULATION> : public BoundaryInterface {
-	static constexpr ProtectionType Protector = TIME_ACCUMULATION;
-	static constexpr const char* format = "\"type\": \"TIME_ACCUMULATION\", \"data\": { \"value\": %s, \"bound\": %s,\"timelimit\": %s }";
-	
-	Boundary(float bound, float time_limit, float frequency, Boundary<float, Protector>*& external_pointer): bound(bound),time_limit(time_limit) ,frequency(frequency), moving_order(frequency*time_limit/100),
+template<typename Type>
+requires(std::is_floating_point_v<Type>)
+struct Boundary<Type, TIME_ACCUMULATION> : public BoundaryInterface {
+	static constexpr ProtectionType Protector = TIME_ACCUMULATION;	
+	Boundary(Type bound, float time_limit, float frequency, Boundary<Type, Protector>*& external_pointer): bound(bound),time_limit(time_limit) ,frequency(frequency), moving_order(frequency*time_limit/100),
```
StefanCostea, line 212: This class is critical, test if it still works after removing specialization

StefanCostea (changes requested):

Src/ST-LIB_HIGH/Protections/Boundary.cpp
g0nz4I0: the map gives me direct access, instead of having to check the type in a switch, I cannot use an enum with the same functionality, also I'd say it is quite readable

Src/ST-LIB_HIGH/Protections/ProtectionManager.cpp
g0nz4I0: it is what it is

Inc/ST-LIB_HIGH/Protections/Boundary.hpp
g0nz4I0: Boundary<ErrorHandler> needs this function, one approach is to implement it only in the Error Handler and then reinterpret_cast the BoundaryInterface* to Boundary<ErroHandler>* and call the method, I think this has some issues when compiling with -fno-rtti,
the other one is to just implement it in the base and override it in the Error Handler one

Inc/ST-LIB_HIGH/Protections/Boundary.hpp
g0nz4I0: ID needs TBD

Inc/ST-LIB_HIGH/Protections/Boundary.hpp
g0nz4I0: just to allocate a max size, I can make the buffer as big as I want.

Inc/ST-LIB_HIGH/Protections/Boundary.hpp
g0nz4I0: will do, but note that the previous implementations were different, one didn't call execute().

ricardochust (approved):

ricardochust:
Remember to add a bit of documentation when its all working

StefanCostea (approved):
LGTM

PR #361, opened 2024-01-10
NTC now uses same approach as other sensors

g0nz4I0:
Minimal changes, so that NTC follows same approach as every other sensor, making it easier and more consistent to use


ricardochust (approved):
i like consistency

AmancioG (approved):

Juanchorosas27 (approved):
Seems good to me

PR #371, opened 2024-03-12
🔥  Mpu manager 🔥 

ricardochust:
The module MPUManager aims to give the user more control on the memory and how it is allocated. For now it is quite simple, it has a few variables to configure the MPU such as using_cache (now the MPU is configured BOTH with and without Ethernet) and has a special function that gets a buffer from a non cacheable space on memory.

This affects multiple modules, which needed to be modified in order to work in all cases, such as Ethernet, SPI, and ADC. The changes control most of the uses for the user, but there is a few changes needed to be made in any project to update it to MPUManager, such as changing the .FLASH and the runes.hpp files to the updated ones. 

Below a more specific list of changes needed:

1. Add MPUConfig declaration on runes `MPUManager::config MPUManager::MPUConfig = {};`

2. Change ADC declaration on modules, as the adc buffers are now declared by the Peripheral constructor instead of needing to receive them in the runes:
```
ADC::Peripheral ADC::peripherals[3] = {
		ADC::Peripheral(&hadc1, adc_buf1, lptim1, init_data1),
		ADC::Peripheral(&hadc2, adc_buf2, lptim2, init_data2),
		ADC::Peripheral(&hadc3, adc_buf3, lptim3, init_data3)
};
```
->
```
ADC::Peripheral ADC::peripherals[3] = {
		ADC::Peripheral(&hadc1, lptim1, init_data1),
		ADC::Peripheral(&hadc2, lptim2, init_data2),
		ADC::Peripheral(&hadc3, lptim3, init_data3)
};
```

3. Add RAM_D3 declaration inside the memory specification area on the FLASH file (jusst below RAM_D2, line 50) `  RAM_D3  (xrw)    : ORIGIN = 0x38000000,   LENGTH = 16K`

4. Add stlib_no_cache_ram_pool declaration on FLASH file (below lwip_sec declaration, line 180):
```
  .stlib_no_cache_ram_pool :
  {
  	. = ABSOLUTE(0x38000000);
  	_no_cached_ram_start = .;
  	
  } >RAM_D3
```

These changes will be added to the template_project, but old projects may want to update by hand instead (such as project of boards)


g0nz4I0:
:warning: I moved this to go into SPI packet casts, as it has almost every single change from that branch also :warning:

PR #372, opened 2024-03-17
Feature/pwm rework

g0nz4I0:
Honestly DualPWM and PWM are very similar, in fact one DualPWM could inherit from PWM and it a coding sense it would make sense, but not in a logical way, same goes for DualPWM containing two PWMs, because of the way Runes are configured these two PWMs would have to be in a way declared in two different maps. So the final approach is to use very similar classes to represent different concepts, also note there is no `virtual` anywhere, I dont think it is needed (why would I ever want a container of different PWMs configurations :) ), also by avoiding `virtual` some overhead might be reduced


ricardochust:
Bug report (yet to fix by me):
Assigning a phase over 40% seems to alter the duty cycle, increasing it over the desired amount. 
Trying to declare a DualPWM as a normal PWM leads into hardFault instead of triggering the ErrorHandler.

Src/HALAL/Services/PWM/DualPWM/DualPWM.cpp
```diff
@@ -76,4 +76,47 @@ void DualPWM::turn_off_negated() {
...
+		sBreakDeadTimeConfig.DeadTime = time /(2 * clock_period_ns) - 64 + 128;
+	}else if(time > 2 * clock_period_ns * 127 && time <= 8 * clock_period_ns * 127){
+		sBreakDeadTimeConfig.DeadTime = time/(8 * clock_period_ns) -32 + 192;
+	}else if(time > 8 * clock_period_ns * 127 && time <=16 * clock_period_ns*127){
+		sBreakDeadTimeConfig.DeadTime = time/(16 * clock_period_ns) -32 + 224;
+	}else{
+		ErrorHandler("Invalid dead time configuration");
+	}
+	//sBreakDeadTimeConfig.LockLevel = 0;
+	//sBreakDeadTimeConfig.BreakState = 1;
```
g0nz4I0, line 116: what is this?

g0nz4I0:
Need to fix having to add a delay or else PWMs dont work

g0nz4I0:
Need to review also InputCapture module, as this is crucial for IMD measurements, all PWM modules should be airtight and fool-proof after merging this PR.
Note that the InputCapture might have to be completely revisited.

1kerProg (approved):

oganigl (approved):

ricardochust:
need to eliminate Half_bridge. Also if you want to check the small issue where the phased_pwm works only on odd number channels that would be cool too

g0nz4I0:
Honestly dead time works fine which is the whole point of this PR, plus also removing the overhead in `virtual` calls to `set_duty_cycle`.

:warning: `Disclaimer:` Phased PWMs may work in unexpected ways, use at own risk :warning: 

With this PR  all boards should be able to use the same ST-LIB branch

PR #373, opened 2024-03-23
SPI workingwithout DMA

g0nz4I0:
Allow to use SPI without having to worry about cache, ideal for interfacing with dummy slaves like ICs such as the IMU or LTC6810.

Note that the last changes on the SPI rework broke the use without DMA, this PR fixes it.

:warning:
 CHANGES ARE NEEDED IN TEMPLATE PROJECT, SPECIFICALLY TO ENABLE/DISABLE DMA for SPI.
Otherwise it will not compile.
:warning:


g0nz4I0:
This merge was tested using a phantom branch combining MPU and this changes. The code was run on the LCU and worked like before.

PR #377, opened 2024-04-13
now spi recovers from a bus collision

ricardochust:
added a recovery in case of a bus collision while recovering from an error on the slave


g0nz4I0:
LGTM, if you still experience issues with SPI consider checking the speed of the pins

g0nz4I0 (approved):

1kerProg (approved):

PR #378, opened 2024-04-24
Keepalives fix

ricardochust:
Need a few changes in lwipopts.h on any project:

1. #define LWIP_TCP_KEEPALIVE 1


g0nz4I0:
LWIP config never dissapoints. but dont use hardcoded values (I know it is on draft, but still)

PR #379, opened 2024-05-08
dont trigger ErrorHandler if SPI transmission fails

g0nz4I0:
ErrorHandler should only occur with serious issues, this doesn't qualify as one.


ricardochust:
Looks good. I don t remember if the HAL_SPI_Transmit had any serious errors though, i will imagine that is not the case. Check it out if you haven t already so we don t end with an unhandled error that disables SPI.

g0nz4I0:
The problem with this is that in the BMS-LIB SPI transmit is called within an Interrupt Handler, therefore sometimes the resource is locked and returns.

In general I think that in Release Builds a lot of ErrorHandler checks could be omitted

1kerProg (approved):

PR #382, opened 2024-05-13
Control block + fix of TimeAcummulation + adc pointer to peripheral data

ricardochust:
Bandaid to temporary solve the problem of Boundary copy on the add_protection macro in combination of the TimeAcummulation calculating out of the Protection handler. The fix makes the TimeAcummulation copies share a pointer to the result of the calculation. No more than one copy of the boundary should ever run the process function. (The check function can be run by any copy without any major trouble).

Added a fixed moving average to work directly with the adc reads (can put any amount of bits in the decimal part of the fixed point data as part of the template argument, and may do so to not lose decimal data). 

Added float versions of the derivator and integrator. Need to be polished on a future update so its part of the template, but for retrocompatibility shake its a different class altogether for now. 

Can get a pointer to the data that the DMA saves on the ADC peripheral to direct access its value without any copy or processing. This also means that any processing (including the shift for the 12 bit ADCs) must be done by the user instead.


jmaralo:
Will be merged with `development_minor_changes`

PR #393, opened 2024-06-16
time config now working on runes

ricardochust:
Lets you configure on runes what timer should be high, mid or global, or even disable them by putting nullptr on the respective pointer (or leaving the set empty for the high timers)

For any project that copied the template before this patch, it is needed to add on runes the next lines, on line 144:`
```
TIM_HandleTypeDef* Time::global_timer = &htim2;
set<TIM_HandleTypeDef*> Time::high_precision_timers = {&htim5, &htim24};
TIM_HandleTypeDef* Time::mid_precision_timer = &htim23;
```

This is the standard configuration used before this patch.


g0nz4I0 (changes requested):
LGTM, but when using for each loops with `auto` use `&` to prevent unnecessary copies, in this case it is not expensive to copy and I guess it works just fine so no big deal.

g0nz4I0 (approved):

PR #412, opened 2024-09-13
Feature/compiles without warnings

g0nz4I0:
Add `-Werror` option to C++ compilation,
additionally added `-Wall` and `-Wpedantic` compiler flags to C++ compilation.

Changed the code to make it compliant with these compiler settings, the changes involve mostly replacing `void*` with `uint8_t*` and fixing `-Wreorder` warnings.

Also removed `[[nodiscard]]` attribute from RingBuffer impl.

Fixed some other warnings.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,9 +68,13 @@ target_compile_options(${LIBRARY} PUBLIC
   -ffunction-sections
   -fdata-sections
   -fno-exceptions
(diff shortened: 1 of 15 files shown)


jmaralo:
Also fixed two more warnings:

* Warning when making a temporal ip address variable that was copied into a memory location allocated using strlen (without taking into acount the null terminator)

PR #414, opened 2024-10-18
Fw-74 Added Pin model for memory structure abstraction

jorgesg82:


Inc/HALALMock/Services/SharedMemory/SharedMemory.hpp
```diff
@@ -9,7 +10,7 @@ class SharedMemory {
     constexpr static uint8_t max_pin_size_bytes{16};
```
jmaralo, line 10: remove this variable altogether or make it initialize to `sizeof(PinModel)`. This size is already encoded in `PinModel`

Inc/HALALMock/Models/PinModel/Pin.hpp
```diff
@@ -78,6 +78,65 @@ enum PinState{
 	ON
 };
 
+enum PinType {
```
jmaralo, line 81:
```suggestion
enum class PinType {
```

jmaralo: regular enums can be mix and matched, for example, I could have another enum somewhere else like:
```c++
enum FDCANMode {
    Loopback,
    External,
};
```
And then somewhere do:
```c++
PinType pin_type = PinMode::DigitalOutput;
if (pin_type == FDCANMode::Loopback) {
    \\ ...
}
```
And this would compile, which is not optimal, because the enum is implicitly casted to a number and compared. `enum class` avoids this issue.

Inc/HALALMock/Models/PinModel/Pin.hpp
```diff
@@ -78,6 +78,65 @@ enum PinState{
...
+	DigitalOutput,
+	DigitalInput,
+	PWM,
+	ADC
+	// TODO: Add more types
+};
+
+struct PinModel {
```
jmaralo, line 89:
```suggestion
struct EmulatedPin {
```
I really don't know how to name this, I don't mind `PinModel`, but at the same time I think we could come up with a better name. Think about it but if you don't find a better name just leave it as is. I think the name I propose in the comment gives a clear impression that is related to the simulator.

Inc/HALALMock/Models/PinModel/Pin.hpp
```diff
@@ -78,6 +78,65 @@ enum PinState{
...
+			// TODO FW-53
+		};
+		struct PWM {
+			// TODO FW-58
+		};
+		struct ADC {
+			// TODO FW-54
+		};
+		// TODO Add more types
+	};
```
jmaralo, line 106: don't know if you tried to compile this, I think the CMake is not ready to compile with the SIM_ON variable in mind, but the way to do this would be more like:
```c++
union {
    struct {
        // TODO FW-52
    } digital_output;

    struct {
        // TODO FW-53
    } digital_input;

    // ...
} pin_data;
```
The other way around you are declaring a named struct and a named union, but not a member of the union or struct if that makes sense

Inc/HALALMock/Services/SharedMemory/SharedMemory.hpp
```diff
@@ -9,7 +10,7 @@ class SharedMemory {
...
 
     // do not use before SharedMemory::start, uninitialized!!!
-    static void *gpio_memory;
+    static PinModel *gpio_memory;
```
jmaralo, line 13: make this private and create a new function to access it, something like:
```c++
static PinModel &get_pin(Pin pin);
```
this prevents code from messing with the base pointer, since once created you just want to access it

Inc/HALALMock/Models/PinModel/Pin.hpp
```diff
@@ -78,6 +78,65 @@ enum PinState{
...
+	{PE14, 78}, {PE15, 79}, {PF0, 80}, {PF1, 81},
+	{PF2, 82}, {PF3, 83}, {PF4, 84}, {PF5, 85},
+	{PF6, 86}, {PF7, 87}, {PF8, 88}, {PF9, 89},
+	{PF10, 90}, {PF11, 91}, {PF12, 92}, {PF13, 93},
+	{PF14, 94}, {PF15, 95}, {PG0, 96}, {PG1, 97},
+	{PG2, 98}, {PG3, 99}, {PG4, 100}, {PG5, 101},
+	{PG6, 102}, {PG7, 103}, {PG8, 104}, {PG9, 105},
+	{PG10, 106}, {PG11, 107}, {PG12, 108}, {PG13, 109},
+	{PG14, 110}, {PG15, 111}, {PH0, 112}, {PH1, 113}
+};
```
jmaralo, line 138: maybe this could go in `SharedMemory.hpp` as it is not closely related with the model, it is just used to access the Shared Memory. It seems like the perfect candidate to be a constant static member in the class `SharedMemory`

jmaralo (changes requested):
A few things to move here and there

Inc/HALALMock/Services/SharedMemory/SharedMemory.hpp
jmaralo: The interface I proposed passed a reference, not a value, maybe you didn't catch it because of the sintax?, as your preference it would be:
```c++
static PinModel& get_pin(Pin pin);
```

jmaralo (approved):

PR #416, opened 2024-10-20
Fw 58

oganigl:
I have create pwm.cpp for the mock, Furthermore I've done the changes in Pin.hpp and SharedMemory requested by Juan in last pull request fw-51

### DIFF
--- a/Inc/HALALMock/Models/PinModel/Pin.hpp
+++ b/Inc/HALALMock/Models/PinModel/Pin.hpp
@@ -88,7 +88,7 @@ enum PinType {
 };
 
 struct PinModel {
-	PinType type; // Always check type before using the union
+	PinType type=PinType::NOT_USED; // Always check type before using the union
 
 	union  {
 		struct DigitalOutput {
--- a/Inc/HALALMock/Models/PinModel/Pin.hpp
+++ b/Inc/HALALMock/Models/PinModel/Pin.hpp
@@ -107,7 +107,7 @@ struct EmulatedPin {
 			// TODO FW-54
 		} ADC;
 		// TODO Add more types
-	}PinData;
+	} PinData;
 };
 
 class Pin {
--- a/Inc/HALALMock/Models/PinModel/Pin.hpp
+++ b/Inc/HALALMock/Models/PinModel/Pin.hpp
@@ -88,7 +88,7 @@ enum class PinType {
 };
 
 struct EmulatedPin {
-	PinType type=PinType::NOT_USED; // Always check type before using the union
+	PinType type = PinType::NOT_USED; //Always check type before using the union
 
 	union  {
 		struct {
--- a/Inc/HALALMock/Models/PinModel/Pin.hpp
+++ b/Inc/HALALMock/Models/PinModel/Pin.hpp
@@ -105,7 +105,7 @@ struct EmulatedPin {
 		}PWM;
 		struct {
 			// TODO FW-54
-		}ADC;
+		} ADC;
 		// TODO Add more types
 	}PinData;
 };
--- a/Inc/HALALMock/Models/PinModel/Pin.hpp
+++ b/Inc/HALALMock/Models/PinModel/Pin.hpp
@@ -78,7 +78,7 @@ enum PinState{
 	ON
 };
 
-enum PinType {
+enum class PinType {
 	NOT_USED,
 	DigitalOutput,
 	DigitalInput,
@@ -87,7 +87,7 @@ enum PinType {
 	// TODO: Add more types
 };
 
-struct PinModel {
+struct EmulatedPin {
 	PinType type=PinType::NOT_USED; // Always check type before using the union
 
 	union  {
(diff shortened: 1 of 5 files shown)


g0nz4I0 (approved):
LGTM (Looks Good To Me) but now that we are within an OS it might be interesting to print more stuff, more info on the console

jmaralo:
# Please DO NOT apply comments targeted to other branches / pull requests in your own branch.

What you should do is, if you need the changes for the comments now and you know how to solve the issues pointed out, first tell the other person you are adding code to solve the comments on their branch because you need it (just in case the other person is about to upload the new code) and then **in the original branch** make a new commit with the changes. Then, once that is done, merge that branch into your own branch and once you are done create the Pull Request relative to the other branch, because now the job is being solved on multiple branches, by multiple people at the same time and the reviewers have to check the changes to the same thing on multiple branches.

In this case, you should have done this:
1. See that you need code from branch `Fw-51` that has an open pull request with comments.
2. Read the changes requested and see that you can apply them on your own.
3. Tell @jorgesg82 that you are fixing some of the comments and if there is any problem with you doing that because it's urgent.
4. Checkout to branch `Fw-51` and write code with the corresponding commits

jmaralo:
Now that it is as it is, no need to change it as it will mess further with the git history, we'll merge the first commit on monday and carry on. But remember to read the comment @oganigl, and also the rest of the subsistem ;)

Src/HALALMock/Services/PWM/PWM/PWM.cpp
```diff
@@ -1,110 +1,63 @@
...
-  if(not is_initialized){
-	ErrorHandler("PWM was not initialized");
-  }
-  if(is_on) return;
-  if(HAL_TIM_PWM_Start(peripheral->handle, channel) != HAL_OK) {
-    ErrorHandler("PWM did not turn on correctly", 0);
-  }
-  is_on = true;
+	if(*(this->frequency)==0){
+		ErrorHandler("You need to set a frequency before turn_on the PWM :%d",*(this->frequency));
```
jmaralo, line 27: If you don't check this condition in set_frequency I wouldn't check it here either, if there is the need for the frequency to be !=0 then the check should be everywhere, otherwise it should be nowhere, as the condition can be easily broken by the user code once started

Src/HALALMock/Services/PWM/PWM/PWM.cpp
```diff
@@ -1,110 +1,63 @@
...
-	{
-		sBreakDeadTimeConfig.DeadTime = time /(2 * clock_period_ns) - 64 + 128;
-	}else if(time > 2 * clock_period_ns * 127 && time <= 8 * clock_period_ns * 127){
-		sBreakDeadTimeConfig.DeadTime = time/(8 * clock_period_ns) -32 + 192;
-	}else if(time > 8 * clock_period_ns * 127 && time <=16 * clock_period_ns*127){
-		sBreakDeadTimeConfig.DeadTime = time/(16 * clock_period_ns) -32 + 224;
-	}else{
-		ErrorHandler("Invalid dead time configuration");
+	if(*(this->is_on)){
+		ErrorHandler("%s","This function can not be called if the PWM is on");
```
jmaralo, line 56: Does the original implementation have this requirement to work, in that case the check should be in both places, otherwise it shouldn't be here.

Src/HALALMock/Services/PWM/PWM/PWM.cpp
```diff
@@ -1,110 +1,63 @@
...
 uint32_t PWM::get_frequency() {
-	return frequency;
+	return *(this->frequency);
 }
 
 float PWM::get_duty_cycle(){
-	return duty_cycle;
+	return *(this->duty_cycle);
 }
 void PWM::set_dead_time(std::chrono::nanoseconds dead_time_ns)
```
jmaralo, line 53: I would add a new variable to the union that represents the dead time in nanoseconds, it is easy and also important to see if the deadtime is being configured

Src/HALALMock/Services/PWM/PWM/PWM.cpp
```diff
@@ -1,110 +1,63 @@
...
+		pin_memory.PinData.PWM.duty_cycle=0.0f;
+		pin_memory.PinData.PWM.is_on=false;
+		//let's point our class variables to the variables from PinModel
+		this->duty_cycle=&(pin_memory.PinData.PWM.duty_cycle);
+		this->frequency=&(pin_memory.PinData.PWM.frequency);
+		this->is_on=&(pin_memory.PinData.PWM.is_on);
+		//default values
+		*(this->duty_cycle)=0.0f;
+		*(this->is_on)=false;
+		*(this->frequency)=0;
```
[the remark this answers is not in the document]
oganigl: Yes true is redundant. I didn't put any frequency due to in the class PWM from the HAL it doesn't so if you want to check frequency before setting it, you will receive a residual value, here as well.

Src/HALALMock/Services/PWM/PWM/PWM.cpp
oganigl: Ok true

Src/HALALMock/Services/PWM/PWM/PWM.cpp
oganigl: Yes it says that you must call the function when the pwm is off

Src/HALALMock/Services/PWM/PWM/PWM.cpp
jmaralo, line 51: I think your implementation is missing an assignment to the duty_cycle_ns variable

Src/HALALMock/Services/SharedMemory/SharedMemory.cpp
```diff
@@ -4,3 +4,14 @@
...
 }
+EmulatedPin &SharedMemory::get_pin(Pin pin){
+    uint8_t offset;
+	auto it=pin_offsets.find(pin);
+	if(it != pin_offsets.end()){
+		offset =it->second;
+	}else{
+		ErrorHandler("Pin %s doesn't exist",pin.to_string());
+	}
+	EmulatedPin *pin_memory=(SharedMemory::gpio_memory)+offset;
```
jmaralo, line 15:
```suggestion
	EmulatedPin *pin_memory = SharedMemory::gpio_memory + offset;
```

Src/HALALMock/Services/SharedMemory/SharedMemory.cpp
```diff
@@ -4,3 +4,14 @@
 void SharedMemory::start() {
     // TODO: create / open shared memory
 }
+EmulatedPin &SharedMemory::get_pin(Pin pin){
+    uint8_t offset;
+	auto it=pin_offsets.find(pin);
+	if(it != pin_offsets.end()){
+		offset =it->second;
```
jmaralo, line 11:
```suggestion
		offset = it -> second;
```

Src/HALALMock/Services/SharedMemory/SharedMemory.cpp
```diff
@@ -4,3 +4,14 @@
 void SharedMemory::start() {
     // TODO: create / open shared memory
 }
+EmulatedPin &SharedMemory::get_pin(Pin pin){
+    uint8_t offset;
+	auto it=pin_offsets.find(pin);
```
jmaralo, line 9:
```suggestion
	auto it = pin_offsets.find(pin);
```

Inc/HALALMock/Models/PinModel/Pin.hpp
```diff
@@ -78,63 +78,35 @@ enum PinState{
...
 };
 
-struct PinModel {
-	PinType type; // Always check type before using the union
+struct EmulatedPin {
+	PinType type=PinType::NOT_USED; // Always check type before using the union
 
-	union PinData {
+	union  {
 		struct DigitalOutput {
```
jmaralo, line 94:
```suggestion
		struct {
```

Inc/HALALMock/Models/PinModel/Pin.hpp
```diff
@@ -78,63 +78,35 @@ enum PinState{
...
-		struct DigitalInput {
+		}DigitalOutput;
+		struct  {
 			// TODO FW-53
-		};
-		struct PWM {
-			// TODO FW-58
-		};
-		struct ADC {
+		}DigitalInput;
```
jmaralo, line 99:
```suggestion
		} DigitalInput;
```

Inc/HALALMock/Models/PinModel/Pin.hpp
```diff
@@ -78,63 +78,35 @@ enum PinState{
...
+		struct  {
+			float duty_cycle;
+			uint32_t frequency;
+			bool is_on;
+		}PWM;
```
jmaralo, line 105:
```suggestion
		} PWM;
```

Inc/HALALMock/Models/PinModel/Pin.hpp
```diff
@@ -78,63 +78,35 @@ enum PinState{
...
+		struct {
 			// TODO FW-54
-		};
+		}ADC;
```
jmaralo, line 108:
```suggestion
		} ADC;
```

Inc/HALALMock/Models/PinModel/Pin.hpp
```diff
@@ -78,63 +78,35 @@ enum PinState{
...
+struct EmulatedPin {
+	PinType type=PinType::NOT_USED; // Always check type before using the union
 
-	union PinData {
+	union  {
 		struct DigitalOutput {
 			// TODO FW-52
-		};
-		struct DigitalInput {
+		}DigitalOutput;
```
jmaralo, line 96:
```suggestion
		} DigitalOutput;
```

Inc/HALALMock/Models/PinModel/Pin.hpp
```diff
@@ -78,63 +78,35 @@ enum PinState{
...
 	DigitalInput,
 	PWM,
 	ADC
 	// TODO: Add more types
 };
 
-struct PinModel {
-	PinType type; // Always check type before using the union
+struct EmulatedPin {
+	PinType type=PinType::NOT_USED; // Always check type before using the union
```
jmaralo, line 91:
```suggestion
	PinType type = PinType::NOT_USED; //Always check type before using the union
```

jmaralo (changes requested):
Incredible styling nitpicks (please try and keep the coding style across files consistent) and a reminder, I think the set_deadtime_ns is not setting the deadtime

jmaralo (approved):

PR #417, opened 2024-10-21
[FW-73] Actualizar modelo PIN de la HALAL Mock

jdmarmen:
Removed all HAL calls and HAL_GPIO_Init struct configurations, void Pin::start() is now empty and does nothing at the moment.


jmaralo (approved):
LGTM (Looks Good To Me), there are still a few references to the types the HAL uses (namely `GPIO_TypeDef`) but that shouldn't be an issue.

PR #421, opened 2024-10-22
Feature/fw 55

oganigl:
I have changed Encoder.hpp and Encoder.cpp to work in the Simulator, the main things are the creation of three privates variables in the class: bool direction, uint32_t count_value, bool is_on.  I linked this variables with the respective EmulatedPin for Pin1 only.


Src/HALALMock/Services/Encoder/Encoder.cpp
```diff
@@ -22,128 +22,73 @@ uint8_t Encoder::inscribe(Pin& pin1, Pin& pin2) {
...
-    return ((timer->handle->Instance->CR1 & 0b10000) >> 4);
+    std::pair<Pin,Pin> pair_pin = Encoder::registered_encoder[id];
+    EmulatedPin &pin1_data = SharedMemory::get_pin(pair_pin.first);
+    return pin1_data.PinData.ENCODER.direction;
 }
 
 uint32_t Encoder::get_initial_counter_value(uint8_t id) {
-    TimerPeripheral* timer = pin_timer_map[registered_encoder[id]];
-    return (timer->handle->Instance->ARR / 2);
+    return 0;
```
jmaralo, line 89: By looking at the original code this should return exactly the half point of the range of the variable you are using to store the counter (e.g. for uint32_t -> UINT32_MAX / 2)

Src/HALALMock/Services/Encoder/Encoder.cpp
```diff
@@ -22,128 +22,73 @@ uint8_t Encoder::inscribe(Pin& pin1, Pin& pin2) {
...
-    if (HAL_TIM_Encoder_Stop(timer->handle, TIM_CHANNEL_ALL) != HAL_OK) {
-        ErrorHandler("Unable to stop encoder in timer %s", timer->name.c_str());
-    }
+    std::pair<Pin,Pin> pair_pin = Encoder::registered_encoder[id];
+    EmulatedPin &pin1_data = SharedMemory::get_pin(pair_pin.first);
+    pin1_data.PinData.ENCODER.is_on = false;
 }
 
-void Encoder::reset(uint8_t id) {
+void Encoder::reset(uint8_t id) {}
```
jmaralo, line 66: set the value to the one the original code is saying, half of the representable range (e.g. for uint32_t -> UINT32_MAX / 2)

jmaralo (changes requested):
two things about the initial value we talked about in the workshop that I realized while looking at the code now

Src/HALALMock/Services/Encoder/Encoder.cpp
oganigl: Yes is a good idea

jmaralo (changes requested):
should be an uint, fix that and is good to go

jmaralo (approved):

g0nz4I0 (approved):

PR #425, opened 2024-10-23
[FW-50] Make the template compile with the mock

jmaralo:
This PR adds support for both compilation for the simulator on the host target (linux support only) or for the MCU. The compilation is configured by passing the `-sim` flag with either "ON" or "OFF" to the python `build.py` script or by calling CMake with `-DSIMULATE=TRUE` or `-DSIMULATE=FALSE`.

The code for the simulator still doesn't compiles because there are still includes from the hal, wich cannot be compiled for the host architecture (it has ASM instructions for the ARM MCU embedded in the code). That should be fixed once all modules are converted, but will take too much for the scope of this PR.

Also the vscode config files have been updated to reflect the changes and the build script now chooses the output path based on the options selected, so that the incremental build is more useful if for example the ST-LIB is first compiled for a project with ethernet and then recompiled for a project without ethernet.


Src/HALALMock/Services/DigitalOutputService/DigitalOutputService.cpp
g0nz4I0, line 1: you have to do something with your linter configuration :)

g0nz4I0 (commented):
still doesnt compile I believe due to ARM inlined assembly

CMakeLists.txt
```diff
@@ -15,6 +13,41 @@ file(GLOB_RECURSE SOURCE_CPP ${CMAKE_SOURCE_DIR}/ *.cpp)
 file(GLOB_RECURSE SOURCE_H ${CMAKE_SOURCE_DIR}/ *.h)
 file(GLOB_RECURSE SOURCE_HPP ${CMAKE_SOURCE_DIR}/ *.hpp)
 
+if(${SIMULATE})
+  include(linux-64.cmake)
+  add_definitions(-DSIM_ON)
```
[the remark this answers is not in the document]
jmaralo: it's literally used in line 51 and the VSCode is consstantly compailing that support for CMake versions < 3.5 is being deprecated, so either we fix the whole file or leave it like this. I think for now it can stay this way

Src/HALALMock/Services/DigitalOutputService/DigitalOutputService.cpp
jmaralo: we can add it to the .vscode/settings.json folder, it is already that way on the CMS-LIB settings, maybe in another PR we can include the settings and apply them to all files there, but now it's a bit hard due to the high amount of PRs

PR #426, opened 2024-10-23
[FW-54] Developed ADC mock

jorgesg82:
Now we have an emulated pin that, in fact, is only a memory pointer which contains the raw value of the simulated ADC. This ADC can be configured in various resolutions: 16, 14, 12, 10 and 8 bits. ADC1 and ADC2 can be configured with all resolutions, while ADC3 only can be configured with <= 12 bit resolution. To manage that, I created an enum that represents all posible resolutions, and is setted comparing with the real resolution of the MCU, which is an uint32_t. Because of that, I don't have implemented 8 bit resolution because its numeric representation depends on a flag that depends on the real MCU, as you can see here:
```
#if defined (ADC_VER_V5_3) || defined(ADC_VER_V5_V90)
#define LL_ADC_RESOLUTION_8B               (ADC_CFGR_RES_2|ADC_CFGR_RES_1 | ADC_CFGR_RES_0) /*!< ADC resolution  8 bits */
#else
#define LL_ADC_RESOLUTION_8B               (ADC_CFGR_RES_2                                ) /*!< ADC resolution  8 bits */
```

For the other side, I  have commented all code that configures parameters related to real peripheral.

Last of all, I have reimplemented HAL calls and getter functions to be used with the mock.

**THIS CODE HAS NOT YET BEEN TESTED, since, at the moment, the STLIB cannot be compiled, and the VMCU has not yet been developed.**


jorgesg82:
I forgot to mention that I have implemented a map that associates an emulated pin with their corresponding instance, as we have in available_instances map with the real pins.

Inc/HALALMock/Services/ADC/ADC.hpp
```diff
@@ -22,24 +23,31 @@ using std::string;
 #define LPTIM3_PERIOD 6875
 
 #define ADC_MAX_VOLTAGE 3.3
-#define MAX_12BIT 4095.0
-#define MAX_16BIT 65535.0
+#define MAX_16BIT 65535.0	// 2^16 - 1
+#define MAX_14BIT 16383.0	// 2^14 - 1
+#define MAX_12BIT 4095.0	// 2^12 - 1
+#define MAX_10BIT 1023.0	// 2^10 - 1
```
jmaralo, line 29: This might be clearer:
```suggestion
#define MAX_16BIT 0b1111111111111111
#define MAX_14BIT 0b0011111111111111
#define MAX_12BIT 0b0000111111111111
#define MAX_10BIT 0b0000001111111111
```
Or by leaving the compiler to do the heavy lifting:
```suggestion
#define MAX_16BIT ((1<<16) - 1)
#define MAX_14BIT ((1<<14) - 1)
#define MAX_12BIT ((1<<12) - 1)
#define MAX_10BIT ((1<<10) - 1)
```

Inc/HALALMock/Services/ADC/ADC.hpp
```diff
@@ -22,24 +23,31 @@ using std::string;
...
+		ADC_RES_12BITS,
+		ADC_RES_10BITS
+	};
+
 	struct InitData {
 	public:
 		ADC_TypeDef* adc;
 		uint32_t resolution;
+		/// @brief Resolution of the emulated ADC translated to an enum
+		ADCResolution emulated_resolution;
```
jmaralo, line 50: don't modify public interfaces, now anything that creates this is invalid because emulated_resolution has not been initialized.

Inc/HALALMock/Services/ADC/ADC.hpp
```diff
@@ -22,24 +23,31 @@ using std::string;
...
 public:
+
+	/// In STM32H723ZG, the ADC1 and ADC2 has 16 bits as their maximum resolution,
+	/// while the ADC3 has 12 bits. Both of them can be configured to has less resolution than its maximum
+	enum ADCResolution {
+		ADC_RES_16BITS,
+		ADC_RES_14BITS,
+		ADC_RES_12BITS,
+		ADC_RES_10BITS
+	};
```
jmaralo, line 43:
```suggestion
	enum class ADCResolution {
		ADC_RES_16BITS,
		ADC_RES_14BITS,
		ADC_RES_12BITS,
		ADC_RES_10BITS
	};
```

Inc/HALALMock/Services/ADC/ADC.hpp
jmaralo, line 50: I think it is more interesting to investigate how the resolution was being used by looking at the code than creating a new enum. Any already existing code won't work with this because they rely on the original resolution, not your resolution implementation

Src/HALALMock/Services/ADC/ADC.cpp
```diff
@@ -17,11 +17,26 @@ unordered_map<uint8_t, ADC::Instance> ADC::active_instances = {};
...
+				case 0x0000000CUL: // 10 bits
+					emulated_resolution = ADCResolution::ADC_RES_10BITS;
+					break;
+			}
+		}
 
 ADC::Peripheral::Peripheral(ADC_HandleTypeDef* handle, LowPowerTimer& timer, InitData& init_data) :
 	handle(handle), timer(timer), init_data(init_data) {
-	dma_data_buffer = (uint16_t*)MPUManager::allocate_non_cached_memory(2*ADC_BUF_LEN);
+	// dma_data_buffer = (uint16_t*)MPUManager::allocate_non_cached_memory(2*ADC_BUF_LEN); Comented because of the mock
```
jmaralo, line 39:
```suggestion
```

Src/HALALMock/Services/ADC/ADC.cpp
```diff
@@ -37,9 +52,16 @@ uint8_t ADC::inscribe(Pin pin) {
 		return 0;
 	}
 
-	Pin::inscribe(pin, ANALOG);
-	active_instances[id_counter] = available_instances[pin];
+	// Emulated pin inscribe
+	EmulatedPin& emulated_pin = SharedMemory::get_pin(pin);
+	if (emulated_pin.type != PinType::NOT_USED) {
+		//ErrorHandler("Pin %s is not available for ADC usage", pin.to_string().c_str());
```
jmaralo, line 58: keep the error handler to mimic original behavior

Src/HALALMock/Services/ADC/ADC.cpp
```diff
@@ -37,9 +52,16 @@ uint8_t ADC::inscribe(Pin pin) {
...
+		printf("Pin %s is not available for ADC usage, is already using as %s", pin.to_string().c_str(), emulated_pin.type);
```
jmaralo, line 59: I guess the error handler will call printf already, so this is not needed

Src/HALALMock/Services/ADC/ADC.cpp
```diff
@@ -77,38 +105,68 @@ void ADC::turn_on(uint8_t id){
 		ErrorHandler("LPTIM - %d - of ADC - %d - did not start correctly", timer.name, peripheral->handle);
 		return;
 	}
+*/ // Commented out because of the mock
```
jmaralo, line 108: Remove commented code if it is not used, the code is already on the original, no need to keep it here

Inc/HALALMock/Models/PinModel/Pin.hpp
```diff
@@ -104,7 +104,7 @@ struct EmulatedPin {
 			std::chrono::nanoseconds dead_time_ns;
 		} PWM;
 		struct {
-			// TODO FW-54
+			uint16_t value;
```
jmaralo, line 107: add a `is_on` variable to see if the ADC has been turned on by the code

Src/HALALMock/Services/ADC/ADC.cpp
```diff
@@ -77,38 +105,68 @@ void ADC::turn_on(uint8_t id){
...
-	uint16_t raw = instance.peripheral->dma_data_buffer[instance.rank];
-	if(instance.peripheral->handle == &hadc3) {
-		return raw / MAX_12BIT * ADC_MAX_VOLTAGE;
+
+	EmulatedPin& emulated_pin = available_emulated_instances[instance];
+	if (emulated_pin.type != PinType::ADC) {
+		printf("Pin %s is not configured to be used for ADC usage. WARNING: it means that this pin \
+				has been reconfigured to be another peripheral WHILE this pin was already configured to be an ADC", emulated_pin);
+		return 0;
 	}
```
jmaralo, line 121: I don't know about this if, the user shouldn't even try to change the pin type mid run, but if you want to keep it change the printf by an error handler

Src/HALALMock/Services/ADC/ADC.cpp
```diff
@@ -77,38 +105,68 @@ void ADC::turn_on(uint8_t id){
...
+	uint16_t raw = emulated_pin.PinData.ADC.value;
+	switch (resolution) {
+		case ADCResolution::ADC_RES_16BITS:
+			return raw / MAX_16BIT * ADC_MAX_VOLTAGE;
+		case ADCResolution::ADC_RES_14BITS:
+			return raw / MAX_14BIT * ADC_MAX_VOLTAGE;
+		case ADCResolution::ADC_RES_12BITS:
+			return raw / MAX_12BIT * ADC_MAX_VOLTAGE;
+		case ADCResolution::ADC_RES_10BITS:
+			return raw / MAX_10BIT * ADC_MAX_VOLTAGE;
```
jmaralo, line 132: parenthesize these expressions and cast to float the MAX_xxBIT just in case someone removes the .0 from the end of the define or another method of representation is used

Src/HALALMock/Services/ADC/ADC.cpp
```diff
@@ -77,38 +105,68 @@ void ADC::turn_on(uint8_t id){
...
 uint16_t ADC::get_int_value(uint8_t id) {
 	Instance& instance = active_instances[id];
-	uint16_t raw = instance.peripheral->dma_data_buffer[instance.rank];
 
-	if(instance.peripheral->handle == &hadc3) {
-		return raw << 4;
+	EmulatedPin& emulated_pin = available_emulated_instances[instance];
+	if (emulated_pin.type != PinType::ADC) {
+		printf("Pin %s is not configured to be used for ADC usage. WARNING: it means that this pin \
+				has been reconfigured to be another peripheral WHILE this pin was already configured to be an ADC", emulated_pin);
```
jmaralo, line 142: turn into error handler

Src/HALALMock/Services/ADC/ADC.cpp
```diff
@@ -77,38 +105,68 @@ void ADC::turn_on(uint8_t id){
...
 
 uint16_t* ADC::get_value_pointer(uint8_t id) {
 	Instance& instance = active_instances[id];
-	return &instance.peripheral->dma_data_buffer[instance.rank];
+	//return &instance.peripheral->dma_data_buffer[instance.rank]; Commented out because of the mock
+	return &available_emulated_instances[instance].PinData.ADC.value;
 }
 
 void ADC::init(Peripheral& peripheral) {
+	/*
```
jmaralo, line 169: remove all dead code

Src/HALALMock/Services/ADC/ADC.cpp
```diff
@@ -77,38 +105,68 @@ void ADC::turn_on(uint8_t id){
...
+	uint16_t raw = emulated_pin.PinData.ADC.value;
+	switch (resolution) {
+		case ADCResolution::ADC_RES_16BITS:
+			return raw;
+		case ADCResolution::ADC_RES_14BITS:
+			return raw << 2;
+		case ADCResolution::ADC_RES_12BITS:
+			return raw << 4;
+		case ADCResolution::ADC_RES_10BITS:
+			return raw << 6;
```
jmaralo, line 158: missing default case if the resolution is not configured correctly

Src/HALALMock/Services/ADC/ADC.cpp
jmaralo, line 132: missing default

Src/HALALMock/Services/ADC/ADC.cpp
```diff
@@ -17,11 +17,26 @@ unordered_map<uint8_t, ADC::Instance> ADC::active_instances = {};
...
+				case 0x00000004UL: // 14 bits
+					emulated_resolution = ADCResolution::ADC_RES_14BITS;
+					break;
+				case 0x00000008UL: // 12 bits
+					emulated_resolution = ADCResolution::ADC_RES_12BITS;
+					break;
+				case 0x0000000CUL: // 10 bits
+					emulated_resolution = ADCResolution::ADC_RES_10BITS;
+					break;
+			}
```
jmaralo, line 34: missing default, call the error handler if the resolution is wrong

jmaralo (changes requested):
Don't modify the public interfaces, and all dead code / commented code that you have remove it, if we need it we have github and the original HALAL code. There are also some places where the code is not returning a value when it should.

Inc/HALALMock/Services/ADC/ADC.hpp
jorgesg82: So how can I implement it? Typing it in private section would work?

Inc/HALALMock/Services/ADC/ADC.hpp
jorgesg82: Why class? Would have any problem if we woudn't use class?

Inc/HALALMock/Services/ADC/ADC.hpp
jorgesg82: Existing code initialises InitData as usual, I'm only using this info to create a new variable where I transform this data to an enum to better manage it, but original resolution remains the same

jorgesg82: If you are talking about for what resolution is used, it is used as a parameter passed to the peripheral for hardware configuration, so is not useful for us as it is.

Inc/HALALMock/Services/ADC/ADC.hpp
```diff
@@ -22,24 +23,31 @@ using std::string;
 #define LPTIM3_PERIOD 6875
 
 #define ADC_MAX_VOLTAGE 3.3
-#define MAX_12BIT 4095.0
-#define MAX_16BIT 65535.0
+#define MAX_16BIT ((1<<16) - 1)
+#define MAX_14BIT ((1<<14) - 1)
+#define MAX_12BIT ((1<<12) - 1)
+#define MAX_10BIT ((1<<10) - 1)
```
g0nz4I0, line 29: careful, this will probably be considered an int by the compiler and it might give you some nasty unexpected results

Inc/HALALMock/Services/ADC/ADC.hpp
g0nz4I0: just using enum puts it in the global namespace, were you to define another enum with the same values you would have a name collision

Src/HALALMock/Services/ADC/ADC.cpp
```diff
@@ -66,107 +92,61 @@ void ADC::turn_on(uint8_t id){
...
 float ADC::get_value(uint8_t id) {
 	Instance& instance = active_instances[id];
-	uint16_t raw = instance.peripheral->dma_data_buffer[instance.rank];
-	if(instance.peripheral->handle == &hadc3) {
-		return raw / MAX_12BIT * ADC_MAX_VOLTAGE;
+
+	EmulatedPin& emulated_pin = available_emulated_instances[instance];
+	if (emulated_pin.type != PinType::ADC) {
+		ErrorHandler("Pin %s is not configured to be used for ADC usage", emulated_pin);
+		return 0;
```
g0nz4I0, line 104: same thing I mentioned in another PR, this is another topic, to add a wrapper around it, for this cyclic and dev only checks

g0nz4I0 (commented):
just implement Juan's feedback

Inc/HALALMock/Services/ADC/ADC.hpp
jmaralo: I made this comment before getting to the bottom of the changed files, basically the same as the comment above

Src/HALALMock/Services/ADC/ADC.cpp
```diff
@@ -17,11 +17,25 @@ unordered_map<uint8_t, ADC::Instance> ADC::active_instances = {};
...
+					break;
+				case 0x00000004UL: // 14 bits
+					emulated_resolution = ADCResolution::ADC_RES_14BITS;
+					break;
+				case 0x00000008UL: // 12 bits
+					emulated_resolution = ADCResolution::ADC_RES_12BITS;
+					break;
+				case 0x0000000CUL: // 10 bits
+					emulated_resolution = ADCResolution::ADC_RES_10BITS;
+					break;
```
jmaralo, line 33: my idea is, instead of here converting to your enum, make each variant of the enum have the corresponding value

Inc/HALALMock/Services/ADC/ADC.hpp
```diff
@@ -22,24 +23,31 @@ using std::string;
...
 public:
+
+	/// In STM32H723ZG, the ADC1 and ADC2 has 16 bits as their maximum resolution,
+	/// while the ADC3 has 12 bits. Both of them can be configured to has less resolution than its maximum
+	enum ADCResolution {
+		ADC_RES_16BITS,
+		ADC_RES_14BITS,
+		ADC_RES_12BITS,
+		ADC_RES_10BITS
+	};
```
jmaralo, line 43: Like so:
```suggestion
	enum ADCResolution : uint32_t {
		ADC_RES_16BITS = 0x00000000,
		ADC_RES_14BITS = 0x00000004,
		ADC_RES_12BITS = 0x00000008,
		ADC_RES_10BITS = 0x0000000C
	};
```

Src/HALALMock/Services/ADC/ADC.cpp
```diff
@@ -66,107 +92,61 @@ void ADC::turn_on(uint8_t id){
...
+	uint16_t raw = emulated_pin.PinData.ADC.value;
+	switch (resolution) {
+		case ADCResolution::ADC_RES_16BITS:
+			return raw / MAX_16BIT * ADC_MAX_VOLTAGE;
+		case ADCResolution::ADC_RES_14BITS:
+			return raw / MAX_14BIT * ADC_MAX_VOLTAGE;
+		case ADCResolution::ADC_RES_12BITS:
+			return raw / MAX_12BIT * ADC_MAX_VOLTAGE;
+		case ADCResolution::ADC_RES_10BITS:
+			return raw / MAX_10BIT * ADC_MAX_VOLTAGE;
```
jmaralo, line 116: and then here and in the other methods do:
```suggestion
	ADCResolution resolution = static_cast<ADCResolution>(instance.peripheral->init_data.resolution);
	uint16_t raw = emulated_pin.PinData.ADC.value;
	switch (resolution) {
		case ADCResolution::ADC_RES_16BITS:
			return raw / MAX_16BIT * ADC_MAX_VOLTAGE;
		case ADCResolution::ADC_RES_14BITS:
			return raw / MAX_14BIT * ADC_MAX_VOLTAGE;
		case ADCResolution::ADC_RES_12BITS:
			return raw / MAX_12BIT * ADC_MAX_VOLTAGE;
		case ADCResolution::ADC_RES_10BITS:
			return raw / MAX_10BIT * ADC_MAX_VOLTAGE;
```

jmaralo (commented):
Basically, this was the idea I had with the enum you proposed, no need to make an initial switch for all possible combinations of `resolution` and you avoid modifying the public interface of the InitData struct.
I don't have any idea in which order the following comments will appear, lets hope that they make sense...

Inc/HALALMock/Services/ADC/ADC.hpp
```diff
@@ -22,137 +23,105 @@ using std::string;
 #define LPTIM3_PERIOD 6875
 
 #define ADC_MAX_VOLTAGE 3.3
-#define MAX_12BIT 4095.0
-#define MAX_16BIT 65535.0
+#define MAX_16BIT ((1 << 16) - 1)
+#define MAX_14BIT ((1 << 14) - 1)
+#define MAX_12BIT ((1 << 12) - 1)
+#define MAX_10BIT ((1 << 10) - 1)
```
jmaralo, line 29:
```suggestion
#define MAX_16BIT 0b1111111111111111
#define MAX_14BIT 0b0011111111111111
#define MAX_12BIT 0b0000111111111111
#define MAX_10BIT 0b0000001111111111
```
same as before but won't overflow an uint16, which could happen by doing (1 << 16)

jmaralo (changes requested):
careful with overflows

jorgesg82 (commented):
All comments resolved

jmaralo (approved):

Inc/HALALMock/Services/ADC/ADC.hpp
```diff
@@ -23,105 +23,105 @@ using std::string;
...
-#define MAX_16BIT ((1<<16) - 1)
-#define MAX_14BIT ((1<<14) - 1)
-#define MAX_12BIT ((1<<12) - 1)
-#define MAX_10BIT ((1<<10) - 1)
-
-
+#define MAX_16BIT 0b1111111111111111
+#define MAX_14BIT 0b0011111111111111
+#define MAX_12BIT 0b0000111111111111
+#define MAX_10BIT 0b0000001111111111
```
g0nz4I0, line 29: my eyes are bleeding, ahahaha , there are gazillions ways to do this and you chose the funniest one ahahaha

usually prefer either 0xFFFFFFFF or C++ equivalents with numeric limits

Inc/HALALMock/Services/ADC/ADC.hpp
```diff
@@ -23,105 +23,105 @@ using std::string;
...
+    enum class ADCResolution : uint32_t {
+        ADC_RES_16BITS = 0x00000000,
+        ADC_RES_14BITS = 0x00000004,
+        ADC_RES_12BITS = 0x00000008,
+        ADC_RES_10BITS = 0x0000000C
+    };
+
+    struct InitData {
+       public:
+        ADC_TypeDef* adc;
```
g0nz4I0, line 45: Not needed

Inc/HALALMock/Services/ADC/ADC.hpp
```diff
@@ -23,105 +23,105 @@ using std::string;
...
+        uint32_t resolution;
+        uint32_t external_trigger;
+        vector<uint32_t> channels;
+        DMA::Stream dma_stream;
```
g0nz4I0, line 49: not needed

Inc/HALALMock/Services/ADC/ADC.hpp
```diff
@@ -23,105 +23,105 @@ using std::string;
...
+        string name;
+
+        InitData() = default;
+        InitData(ADC_TypeDef* adc, uint32_t resolution,
+                 uint32_t external_trigger, vector<uint32_t>& channels,
+                 DMA::Stream dma_stream, string name);
```
g0nz4I0, line 55: not needed

Inc/HALALMock/Services/ADC/ADC.hpp
```diff
@@ -23,105 +23,105 @@ using std::string;
...
+    };
+
+    class Peripheral {
+       public:
+        ADC_HandleTypeDef* handle;
+        uint16_t* dma_data_buffer;
+        LowPowerTimer timer;
+        InitData init_data;
```
g0nz4I0, line 63: not needed

g0nz4I0 (changes requested):
The ADC should be super simple, you dont have to worry about channels/ranks, maybe only resolution but you can probably come up with an easy/trick solution

jmaralo (approved):

PR #427, opened 2024-10-23
FW-78 HALAL Mock Time

StefanCostea:
Added Mock for time.
Support for time speed up.
Based in std::thread.
Thread safe, all critical methods use locks.
There is a lag calculation mechanism to detect wether the machine can keep up with the simulation, currently not working as expected.
Addition of Test folder, including base test case and test for the lag mechanism.
The test for the lag mechanism is supposed to search for the maximum speed up the host machine can safely handle.
TODO: repair lag detection mechanism
TODO: Fill Error Handler methods


StefanCostea:
Rebase this PR to the correct branch for merge queue to develop

Inc/HALALMock/Services/Time/Time.hpp
```diff
@@ -1,148 +1,67 @@
...
-	/**
-	 * @brief Cancels a timeout by derigstering the alarm bound to it
-	 * 
-	 * @param id The id of the timeout to cancel
-	 */
-	static void cancel_timeout(uint8_t id);
+private:
+    using RealClock = std::chrono::steady_clock;
+    using RealTimePoint = RealClock::time_point;
+    using SimDuration = std::chrono::nanoseconds;
```
g0nz4I0, line 26: nice flex

Src/HALALMock/Services/Time/Time.cpp
```diff
@@ -1,445 +1,168 @@
...
-        } else {
-            hal_enable_timer(global_timer);
-            Time::init_timer(
-                timer32_by_timer32handler_map[global_timer], global_timer, 0,
-                HIGH_PRECISION_MAX_ARR,
-                timer32interrupt_by_timer32handler_map[global_timer]);
-            HAL_TIM_Base_Start_IT(global_timer);
-        }
+void Time::start(double simulation_speed) {
+    std::lock_guard<std::mutex> lock(mutex);
```
g0nz4I0, line 26: RAII, I like this one

Src/HALALMock/Services/Time/Time.cpp
```diff
@@ -1,445 +1,168 @@
...
-        Time::mid_precision_registered = true;
-    }
-
-    alarm.alarm = func;
-    Time::mid_precision_alarms_by_id[mid_precision_ids] = alarm;
-    NVIC_EnableIRQ(timer32interrupt_by_timer32handler_map[mid_precision_timer]);
-
-    return mid_precision_ids;
+uint8_t Time::register_low_precision_alarm(uint32_t period_in_ms, std::function<void()> func) {
+    return register_high_precision_alarm(period_in_ms * 1000, func); //There is no hw limitation that does not allow this
```
g0nz4I0, line 81: you mean regarding Timers as they are currently configured in STLIB right?

Src/HALALMock/Services/Time/Time.cpp
```diff
@@ -1,445 +1,168 @@
...
-}
 
-void Time::low_precision_timer_callback() {
-    for (auto& pair : Time::low_precision_alarms_by_id) {
-        Time::Alarm& alarm = pair.second;
-        if (alarm.is_on == false || alarm.offset == Time::low_precision_tick)
-            continue;
-        if ((Time::low_precision_tick - alarm.offset) % alarm.period == 0) {
-            alarm.alarm();
+        for (auto it = alarms.begin(); it != alarms.end();) {
```
g0nz4I0, line 149: nitpick use structure binding for clearer naming

Src/HALALMock/Services/Time/Time.cpp
```diff
@@ -1,445 +1,168 @@
...
+            if (it->second.active && it->second.next_trigger <= simulation_time) {
```
g0nz4I0, line 150: I assume the `<=` is to account for drift? as ideally the timer will only awake after having waiter_for the diferente of the expected duration for next trigger?

g0nz4I0 (approved):
Some stuff I'm not able to fully grasp :), I get the general idea and if the test passes then it should be good

Src/HALALMock/Services/Time/Time.cpp
```diff
@@ -1,445 +1,168 @@
...
  * Time.cpp
  *
- *  Created on: 30 oct. 2022
- *      Author: Dani
+ *  Created on: 23 oct. 2024
+ *      Author: Stefan Costea
  */
 
-#include "Time/Time.hpp"
+#include "Inc/HALALMock/Services/Time/Time.hpp"
```
jmaralo, line 8: This should be just
```suggestion
#include "HALALMock/Services/Time/Time.hpp"
```

Src/HALALMock/Services/Time/Time.cpp
```diff
@@ -1,445 +1,168 @@
...
-            high_precision_timers.erase(timer);
-        } else {
-            hal_enable_timer(timer);
-            Time::init_timer(timer32_by_timer32handler_map[timer], timer, 0,
-                             HIGH_PRECISION_MAX_ARR,
-                             timer32interrupt_by_timer32handler_map[timer]);
-            Time::available_high_precision_timers.push(timer);
-        }
+    if (simulation_speed <= 0.0) {
+        //TODO: ErrorHandler("Simulation speed must be positive");
```
jmaralo, line 31: Would already add an error handler here with the TODO to maybe change it in the future for the simulator-specific logging system we use

jmaralo (commented):
Mostly what @g0nz4I0 said, just two things to say

Src/HALALMock/Services/Time/Time.cpp
StefanCostea: Yes, its to account for drift

Src/HALALMock/Services/Time/Time.cpp
StefanCostea: In HALAL there is three levels of timers because of hardware limitations, in the mock they do not exist, i just use high precision for everything and add extra methods to comply with the interface

Src/HALALMock/Services/Time/Time.cpp
StefanCostea: what do you mean?

Src/HALALMock/Services/Time/Time.cpp
StefanCostea: That path did not work for me, there should be a setup.env file in the workspace that sets all the necessary paths to support that and other setup actions

Src/HALALMock/Services/Time/Time.cpp
g0nz4I0:
```c++
for( auto& [id, alarm] : alarms){
    if(alarm.active && alarm.next_trigger <= simulation_time){
       // more code here
    }
}
```

g0nz4I0 (approved):

Src/HALALMock/Services/Time/Time.cpp
StefanCostea: cant do that, the reason i am doing it with iterators is because i am deleting items as i iterate through the map, doing that like that there is no way of skipping an item when erased

Src/HALALMock/Services/Time/Time.cpp
```diff
@@ -1,445 +1,168 @@
...
-void Time::low_precision_timer_callback() {
-    for (auto& pair : Time::low_precision_alarms_by_id) {
-        Time::Alarm& alarm = pair.second;
-        if (alarm.is_on == false || alarm.offset == Time::low_precision_tick)
-            continue;
-        if ((Time::low_precision_tick - alarm.offset) % alarm.period == 0) {
-            alarm.alarm();
+        for (auto it = alarms.begin(); it != alarms.end();) {
+            if (it->second.active && it->second.next_trigger <= simulation_time) {
+                auto alarm = it->second;
```
g0nz4I0, line 151: also, why the copy?

Src/HALALMock/Services/Time/Time.cpp
g0nz4I0: Completely right, I though you could get away with map::erase(const Key& key) but it segfault

jmaralo (approved):

PR #428, opened 2024-10-24
[FW-56] GPIO EXTI

Cantonplas:
Pin.hpp: Defined the EXTI type in EmulatedPin and created its variables for the shared memory. 
EXTI.cpp: Modified it to try to allow it to be used by the python script to make it detect a trigger and send an interrupt that will call a callback function.

Incomplete things: Can´t select a priority for the interrupt signal.


Src/HALALMock/Services/EXTI/EXTI.cpp
```diff
@@ -65,5 +69,10 @@ bool ExternalInterrupt::get_pin_value(uint8_t id) {
 	}
 
 	Pin& pin = service_ids[id];
-	return HAL_GPIO_ReadPin(GPIO_PORT, pin.gpio_pin);
+	EmulatedPin &pin_data = SharedMemory::get_pin(pin);
+	if(pin_data.type != PinType::EXTIPin) {
+		ErrorHandler("ID %d is not registered as a EXTIPin",id);
+		return;
+	}
```
g0nz4I0, line 76: I know this is probably how it is coded in the "normal" version of the STLIB, it would be interesting to make a wrapper around it so this check is only performed in debug builds / can be switched on or off at compile time.

Src/HALALMock/Services/EXTI/EXTI.cpp
```diff
@@ -26,25 +26,23 @@ uint8_t ExternalInterrupt::inscribe(Pin& pin, function<void()>&& action, TRIGGER
...
 
+	(pin_data.PinData.EXTIPin.trigger_mode) = trigger;
+
 	service_ids[id_counter] = pin;
 	instances[pin.gpio_pin].action = action;
 
 	return id_counter++;
 }
-
+//TODO: assigne priority on the emulated pin
```
g0nz4I0, line 42: Don't worry too about priority of interrupts for now, we dont use them in the STLIB, (although we should) .

In case we want to actually use priorities we could make a global queue of signals

Inc/HALALMock/Models/PinModel/Pin.hpp
```diff
@@ -106,6 +107,13 @@ struct EmulatedPin {
 		struct {
 			// TODO FW-54
 		} ADC;
+		struct{
+			uint32_t priority=0;
+			bool is_on;
+			bool trigger_signal;
+			TRIGGER trigger_mode;
+			
+		} EXTIPin;
```
g0nz4I0, line 116: General question, does this compile?, like the whole union

g0nz4I0 (approved):
Should work

jmaralo (approved):

PR #429, opened 2024-10-25
Added Input Capture Mock Implementation

Ikubimu:


Src/HALALMock/Services/InputCapture/InputCapture.cpp
```diff
@@ -18,15 +18,12 @@ static map<uint32_t, uint32_t> channel_dict = {
...
-	channel_falling(channel_falling)
+InputCapture::Instance::Instance(Pin& pin, void* peripheral, uint32_t channel_rising, uint32_t channel_falling) :
+	pin(pin)
 	{
-		frequency = 0;
-		duty_cycle = 0;
-
+		EmulatedPin& sim_pin = SharedMemory::get_pin(pin);
+		duty_cycle = &(sim_pin.PinData.InputCapture.duty_cycle);
+		frequency = &(sim_pin.PinData.InputCapture.frequency);
```
jmaralo, line 26: set to zero here, the initialization of duty_cycle and frequency has to take place here

jmaralo (changes requested):
Just a thing on the initialization of variables, also I don't know how will the changes to the pubic interface will affect already existing user code, but we can worry about that later when we start integrating everything with actual user code

jmaralo (approved):

Inc/HALALMock/Services/InputCapture/InputCapture.hpp
jorgesg82, line 1: I wouldn't modify the public interface. If you don't use these variables, leave it as is, as dead code, just in case the user code does something nasty with it. Leaving it as is shouldn't disturb you

jorgesg82 (approved):

Inc/HALALMock/Services/InputCapture/InputCapture.hpp
```diff
@@ -2,41 +2,40 @@
...
-		uint32_t frequency;
-		uint8_t duty_cycle;
+   public:
+    class Instance {
+       public:
+        uint8_t id;
+        Pin pin;
+        uint32_t* frequency;
+        uint8_t* duty_cycle;
+        bool is_active;
```
jmaralo, line 23: didn't saw this until Jorge pointed it out, you are changing the public interface, you can add these variables as new members instead of changing the original because you don't know if the user will modify/access them. Return the implementation to how it was before and add these pointers as extra members

jmaralo (approved):

PR #430, opened 2024-10-25
[FW-72] Create SHM zone on SharedMemory::start

oganigl:
I have added the function start() in sharedMemory, to create the amount of shared memory enough for all the gpio, I added some variables as the name, and the size in the .hpp


jmaralo (changes requested):
From now on I will also review PR names, this should be something like `[FW-72] Create SHM zone on SharedMemory::start`, not the branch name, we talked about this in the last meeting.

Src/HALALMock/Services/SharedMemory/SharedMemory.cpp
```diff
@@ -1,8 +1,34 @@
...
+#include <sys/stat.h>
+#include <sys/mman.h>
+#include <unistd.h>
 void SharedMemory::start() {
-    // TODO: create / open shared memory
+    //Create GPIO_Memory
+	const char* 
+	int shm_gpio_fd;
+	//create shared memory object
+	shm_gpio_fd = shm_open(gpio_memory_name, O_CREAT | O_RDWR,0666);
```
jorgesg82, line 14: '0666' mode allows to everybody to read and write on this file descriptor, it wouldn't better to use 0660? It would allow to propietary and group to read and write, but rest of users couldn't do anything
```suggestion
	shm_gpio_fd = shm_open(gpio_memory_name, O_CREAT | O_RDWR,0660);
```

jorgesg82 (changes requested):

Src/HALALMock/Services/SharedMemory/SharedMemory.cpp
oganigl: Yes, it's be a good Idea, for safety reasons.

Inc/HALALMock/Services/SharedMemory/SharedMemory.hpp
```diff
@@ -3,43 +3,45 @@
...
 #include "ErrorHandler/ErrorHandler.hpp"
+
 class SharedMemory {
    public:
-    // Number of pins per GPIO port
-    constexpr static uint8_t pins_per_port = 16;
-
+	constexpr static uint8_t total_pins= 114;
+	constexpr int gpio_memory_size = total_pins * sizeof(EmulatedPin);
+	constexpr char *gpio_memory_name = "/Shared_Memory_GPIO";
```
g0nz4I0, line 11: change the name, otherwisw we cannot launch several instances

Src/HALALMock/Services/SharedMemory/SharedMemory.cpp
```diff
@@ -1,8 +1,34 @@
...
+//includes to create the shared Memory in Posix
+#include <fcntl.h>
+#include <sys/shm.h>
+#include <sys/stat.h>
+#include <sys/mman.h>
+#include <unistd.h>
 void SharedMemory::start() {
-    // TODO: create / open shared memory
+    //Create GPIO_Memory
+	const char*
```
g0nz4I0, line 11: ??

g0nz4I0 (changes requested):
Dont hardcode shm name, dont use ErrorHandler for showing errors when creating the Shared Memory

Inc/HALALMock/Services/SharedMemory/SharedMemory.hpp
jmaralo, line 11: make this an extern variable that the user sets in a config file

Src/HALALMock/Services/SharedMemory/SharedMemory.cpp
jmaralo: what gonzalo said "??", this line seems to be a leftover from something you were doing

Src/HALALMock/Services/SharedMemory/SharedMemory.cpp
```diff
@@ -1,8 +1,34 @@
...
+	//configure the size of the shared memory object
+	if(ftruncate(shm_gpio_fd,gpio_memory_size) == -1){
+		ErrorHandler("%s","Error to asssign memory to the Shared Memory");
+		close(shm_gpio_fd);
+		return;
+	}
+	//point gpio_memory to the beginning of shared_memory
+	gpio_memory = mmap(0, gpio_memory_size, PROT_WRITE | PROT_READ, MAP_SHARED, shm_gpio_fd, 0);
+	if(gpio_memory == MAP_FAILED){
+		ErrorHandler("%s", "Error al mapear la memoria compartida");
```
jmaralo, line 28:
```suggestion
		ErrorHandler("%s", "Error mapping Shared Memory");
```

jmaralo (changes requested):
Main thing is the name

g0nz4I0 (approved):

jorgesg82 (approved):

PR #431, opened 2024-10-26
New implementation for the ringBuffer by index

Ikubimu:


jmaralo (commented):
Please add a description in your PRs, is useful when reviewing and to warn about certain known bugs or issues

jmaralo (approved):
Looks good, but there is already a half working implementation of a ring buffer in `Inc/C++Utilities/RingBuffer.hpp`, the idea was to fix that implementation, but if this works I won't complain (at least not until we need to use another ring buffer other than this :))

Ikubimu:
Ok I didnt know, I ll take a look at

g0nz4I0 (changes requested):
Looks good, but could you move the implementation to the existing RingBuffer to replace it? Is is in  `Inc/C++Utilities/RingBuffer.hpp`, also this way we can test it more easily

Src/ST-LIB_LOW/Sensors/EncoderSensor/EncoderSensor.cpp
```diff
@@ -1,7 +1,7 @@
 #include "Sensors/EncoderSensor/EncoderSensor.hpp"
 
 EncoderSensor::EncoderSensor(Pin pin1, Pin pin2, double *position, double* direction, double *speed, double *acceleration)
-: position(position), direction(direction), speed(speed), acceleration(acceleration){
+: position(position), direction(direction), speed(speed), acceleration(acceleration), idx_buff(0){
```
g0nz4I0, line 4: TIP: when you declare it in the class you can do: `uint32_t idx_buff{0};` and that will initialize it, without having to explicitly list it out in the constructor

Ikubimu:
Just  a thing, the actual RingBuffer implementation manage the buffer as a consumer-producer , this means that can't overwrite on the buffer. It doesn't make sense to me because the speetech is continuously reading data. Should a new template model be done to implement that way or something?

g0nz4I0:
What do you mean as a consumer-producer, I see that push fails if its full, therefore you should call `pop` before, but feel free to reimplement it

Ikubimu:
Encoder sensor obtains the last measure by pointer, then doesn't care the about the buffer. And the buffer fails if is full, then it would be all the time pop and push and I see innecesary

Ikubimu:
If Encoder is always measuring,  shouldn't it be able to update the buffer ignoring if is full or not?

jmaralo (approved):

g0nz4I0 (approved):
Easy way of seeing if it works, if it doesnt we just revert

PR #432, opened 2024-10-26
[FW-82] Mocked SPI

jorgesg82:
To simulate SPI communication, I used TCP sockets. One socket is created per SPI peripheral registered. Through this socket, transmission and reception are simulated.

The socket creation and configuration is made when an SPI peripheral is registered, binding to an specific local port starting from port 2000. The MCU we use has six SPI peripherals, so ports 2000 to 2005 can be used to SPI communication. At this moment, socket already tries to connect to the simulator.

When a transmission or reception must be handled, socket only sends or receives data through the connection already created when the SPI peripheral was registered.

All peripheral configuration has been removed, as we are not working with real peripherals.

The transmission and reception callback is called when transmission and reception through TCP socket has been successful


Inc/C++Utilities/CppImports.hpp
jorgesg82, line 1: I need this if I'm going to run this code in my mac. Sorry

Src/HALALMock/Services/Communication/SPI/SPI.cpp
```diff
@@ -62,6 +65,35 @@ uint8_t SPI::inscribe(SPI::Peripheral& spi) {
     SPI::registered_spi[id] = spi_instance;
     SPI::registered_spi_by_handler[spi_instance->hspi] = spi_instance;
 
+    // Create socket for this SPI peripheral
+    int spi_socket = socket(AF_INET, SOCK_STREAM, 0);
+
+    // Bind this socket to the local port 200x
+    sockaddr_in local_address;
+    local_address.sin_family = AF_INET;
+    local_address.sin_port = htons(2000 + id_counter);
```
g0nz4I0, line 74: dont use magic numbers, do something like `#define SPI_PORT_BASE 2000 `and then use like `htons(SPI_PORT_BASE + id_counter)`

Src/HALALMock/Services/Communication/SPI/SPI.cpp
```diff
@@ -104,54 +134,15 @@ bool SPI::transmit(uint8_t id, span<uint8_t> data) {
...
-        default:
-            ErrorHandler(
-                "Error while transmiting and receiving with spi DMA. Errorcode "
-                "%u",
-                (uint8_t)errorcode);
-            return false;
-            break;
-    }
+bool SPI::transmit_DMA(uint8_t id, span<uint8_t> data) {
+    return SPI::transmit(id, data);
```
g0nz4I0, line 145: add a comment explaining that in simulator mode no DMA is used

Src/HALALMock/Services/Communication/SPI/SPI.cpp
```diff
@@ -512,19 +436,22 @@ void HAL_SPI_ErrorCallback(SPI_HandleTypeDef* hspi) {
 void SPI::spi_communicate_order_data(SPI::Instance* spi, uint8_t* value_to_send,
                                      uint8_t* value_to_receive,
                                      uint16_t size_to_send) {
-    HAL_SPI_TransmitReceive_DMA(spi->hspi, value_to_send, value_to_receive,
-                                size_to_send);
-}
+    // Search for the id of the SPI instance
+    uint8_t id;
+    for (auto iter : registered_spi) {
```
g0nz4I0, line 441: use structure bindings

Src/HALALMock/Services/Communication/SPI/SPI.cpp
```diff
@@ -189,27 +164,13 @@ bool SPI::transmit_and_receive(uint8_t id, span<uint8_t> command_data,
...
-            break;
-        default:
-            ErrorHandler(
-                "Error while transmiting and receiving with spi DMA. Errorcode "
-                "%u",
-                (uint8_t)errorcode);
-            return false;
-            break;
+    if(SPI::transmit(id, command_data) && SPI::receive(id, receive_data)) {
+        HAL_SPI_TxRxCpltCallback(SPI::registered_spi[id]->hspi);
```
g0nz4I0, line 168: should this be here? it is introducing the HAL Typedef, I don't know the policy on this

jorgesg82: Other posibility is renaming callback name, but, although this function is only used by HAL, is a public method and I would be changing a public interface.

Src/HALALMock/Services/Communication/SPI/SPI.cpp
jorgesg82: What do you mean?

Src/HALALMock/Services/Communication/SPI/SPI.cpp
jmaralo: I think it is possible that user code uses this callback underneath, we can leave it like is for now and come back later if it doesn't compile

Inc/C++Utilities/CppImports.hpp
```diff
@@ -16,7 +16,11 @@
 #include <array>
 #include <bitset>
 #include <ctype.h>
-#include <malloc.h>
+#ifdef __APPLE__
+#include <stdlib.h>  // macOS
+#else
+#include <malloc.h>  // Linux/Unix
+#endif
```
jmaralo, line 23: This file is not taken into account in the mock, also protect this by using SIM_ON
```suggestion
#ifdef SIM_ON
#ifdef __APPLE__
#include <stdlib.h>  // macOS
#else
#include <malloc.h>  // Linux/Unix
#endif
#else
#include <malloc.h>
#endif
```
Just in case for some reason when compiling from a mac to the MCU the flag __APPLE__ is set for some unknown reason

Src/HALALMock/Services/Communication/SPI/SPI.cpp
```diff
@@ -62,6 +63,37 @@ uint8_t SPI::inscribe(SPI::Peripheral& spi) {
...
+    server_address.sin_family = AF_INET;
+    server_address.sin_port = htons(2000 + id_counter);
+    server_address.sin_addr.s_addr = INADDR_ANY;
+
+    if (connect(spi_socket, (struct sockaddr*)&server_address,
+                sizeof(server_address)) < 0) {
+        ErrorHandler("Connect error: %s", strerror(errno));
+        close(spi_socket);
+        return 0;
+    }
```
jmaralo, line 93: this should be a server, I believe connect blocks until the connection is made, so if the user just wants one SPI instance but the code declares 6, the first will connect but the other five ones will wait for a connection that won't be made

Src/HALALMock/Services/Communication/SPI/SPI.cpp
```diff
@@ -189,26 +163,11 @@ bool SPI::transmit_and_receive(uint8_t id, span<uint8_t> command_data,
...
-            break;
-        default:
-            ErrorHandler(
-                "Error while transmiting and receiving with spi DMA. Errorcode "
-                "%u",
-                (uint8_t)errorcode);
-            return false;
-            break;
+    if (SPI::transmit(id, command_data) && SPI::receive(id, receive_data)) {
+        HAL_SPI_TxRxCpltCallback(SPI::registered_spi[id]->hspi);
```
jmaralo, line 167: you are not calling a method like this one when just transmitting or receiving, so why do it here, if you want to call the HAL callbacks you should also do it in the transmit and recieve methods.

Src/HALALMock/Services/Communication/SPI/SPI.cpp
jmaralo: I'm guessing [this](https://en.cppreference.com/w/cpp/language/structured_binding)

Src/HALALMock/Services/Communication/SPI/SPI.cpp
```diff
@@ -512,19 +434,22 @@ void HAL_SPI_ErrorCallback(SPI_HandleTypeDef* hspi) {
...
+                              span<uint8_t>(value_to_receive, size_to_send));
 }
 
-void SPI::turn_off_chip_select(SPI::Instance* spi) {
-    HAL_GPIO_WritePin(spi->SS->port, spi->SS->gpio_pin,
-                      (GPIO_PinState)PinState::OFF);
-}
+void SPI::turn_on_chip_select(SPI::Instance* spi) {}
+
+void SPI::turn_off_chip_select(SPI::Instance* spi) {}
```
jmaralo, line 452: This should update the pin in memory, it is easy to simulate

jmaralo (changes requested):
Main thing is that the SPI peripherals of the simulator should act as servers because from what I know, clients will block (aka stop) the process until the connection with the server is made, but there are multiple SPI instances when we will just use one or two and the remaining, unused, SPI instances will block forever because nobody will try to connect to them

Src/HALALMock/Services/Communication/SPI/SPI.cpp
jorgesg82: But a socket only creates and tries to connect when a SPI peripheral REGISTERS, that means when the user is going to use it. Not the six sockets is going to create unless the user want to use them all

Src/HALALMock/Services/Communication/SPI/SPI.cpp
jorgesg82: But transmit and receive methods are empty

Src/HALALMock/Services/Communication/SPI/SPI.cpp
```diff
@@ -1,16 +1,16 @@
...
+#include <netinet/in.h>
+#include <sys/socket.h>
+#include <unistd.h>
+
 #include "HALALMock/Models/MPUManager/MPUManager.hpp"
+#include "HALALMock/Services/SharedMemory/SharedMemory.hpp"
 
 #ifdef HAL_SPI_MODULE_ENABLED
 
+#define SPI_PORT_BASE 2000
```
StefanCostea, line 12: Now that all the services are going to start using sockets for everything i see just right to have something that unifies all the ports somewhere, something like a piece of code that has all the hardcoded ports for each of the services and checks for possible collisions or problems with the port already being used by the OS

Src/HALALMock/Services/Communication/SPI/SPI.cpp
jorgesg82: Done

Src/HALALMock/Services/Communication/SPI/SPI.cpp
g0nz4I0: I believe a possible solution is: if the peripheral is MASTER_MODE, launch a server, otherwise just become a client.

and I think @jorgesg82 is rith, you only open the socket when calling spi::inscribe, so that should be fine

StefanCostea:
Add Tests, now that everything runs on OS should be mandatory to have tests related to every module

Src/HALALMock/Services/Communication/SPI/SPI.cpp
g0nz4I0: Dont use the callbacks, ideally we will drop all of the HAL codebase, so we cannot be introducing dependencies. The callback basically implements a state machine using the status of the SPI bus, we should be abble to bypass it

Inc/HALALMock/Models/PinModel/Pin.hpp
```diff
@@ -6,156 +6,158 @@
...
-	DigitalOutput,
-	DigitalInput,
-	PWM,
-	DualPWM,
-	ADC
-	// TODO: Add more types
+    NOT_USED,
+    DigitalOutput,
+    DigitalInput,
+    PWM,
```
StefanCostea, line 81: remove this changes form this branch, this PR should just include what it adds, from my perspective if you have dependency on this merge this PR into the branch that adds this to develop and when that branch gets merged rebase the PR to develop

StefanCostea (changes requested):

Src/HALALMock/Services/Communication/SPI/SPI.cpp
StefanCostea: At least having a sys_config file with all of this parameters somewhere, there are going to be lots of parameters that it would be helpful to have a single place to change everything

Src/HALALMock/Services/Communication/SPI/SPI.cpp
jorgesg82: To implement it, I created a bool in pin structure that updates from here

Src/HALALMock/Services/Communication/SPI/SPI.cpp
jorgesg82: I recoded this part.

Now a peripheral that acts as a master creates a server socket, while a peripheral that acts as a salve creates a client socket.

Master peripherals use a pin as a Slave Select, which turns on when the peripheral start. I used this to tell the simulator when to connect, putting a bool true in the emulated pin.

The ST-LIB is coded to connect only one slave to one master, so when the peripheral starts the server only waits to ONE slave to connect

Inc/HALALMock/Models/PinModel/Pin.hpp
jorgesg82: But to add SPI mock I had to create a new type in emulated pin. It is directly related to this PR

Src/HALALMock/Services/Communication/SPI/SPI.cpp
jorgesg82: Transmit callback and receive callback is declared but empty, only transmission-and-reception callback is doing stuff, so I leave it as it is.

Src/HALALMock/Services/Communication/SPI/SPI.cpp
jorgesg82: This is not strictly related to this PR, is more a structure matter. Please, discuss with @g0nz4I0 and @jmaralo

Inc/HALALMock/Models/PinModel/Pin.hpp
g0nz4I0: it is just a linter issue :))

Src/HALALMock/Services/Communication/SPI/SPI.cpp
g0nz4I0: Ideally we could have a small GUI to configure all the settings, and then generate a header file for C++ and another generated one for the python tests, something like the runes but more aimed to the simulator.

Inc/HALALMock/Models/PinModel/Pin.hpp
StefanCostea: But the whole file is changed

StefanCostea (dismissed):

Src/HALALMock/Services/Communication/SPI/SPI.cpp
g0nz4I0: I don't see the logic you mention, the SPI master should launch a server and wait for outside connections, maybe from another sensor/ from the Python generator, or another board (LCU-BCU).

If it's a slave it should block until successfully connecting to a master

g0nz4I0: there is no `listen()` call?

Src/HALALMock/Services/Communication/SPI/SPI.cpp
```diff
@@ -31,30 +31,77 @@ uint8_t SPI::inscribe(SPI::Peripheral& spi) {
...
     uint8_t id = SPI::id_counter++;
+
+    // Create socket for this SPI peripheral
+    int spi_socket = socket(AF_INET, SOCK_STREAM, 0);
+
+    // Bind this socket to the local port 200x
+    sockaddr_in local_address;
+    local_address.sin_family = AF_INET;
+    local_address.sin_port = htons(SPI_PORT_BASE + id_counter);
+    local_address.sin_addr.s_addr = INADDR_ANY;
```
g0nz4I0, line 58: address should be fixed, otherwise how are clients going to connect?

g0nz4I0 (changes requested):

Inc/HALALMock/Models/PinModel/Pin.hpp
jorgesg82: Someone made a merge in development branch messing up everything and now seems that this file is totally changed, but is not my fault!

Src/HALALMock/Services/Communication/SPI/SPI.cpp
jorgesg82: But it is assumed that clients will be in the same host, isn't it?

g0nz4I0: I looked into the documentation for INADDR_ANY, and yes that should work further info, [ understanding-inaddr-any-for-socket-programming](https://stackoverflow.com/questions/16508685/understanding-inaddr-any-for-socket-programming)

They will be in the same host yes, but maybe they have differents IPs, but INADDR_ANY bounds to **ALL** addresses so it works fine

Src/HALALMock/Services/Communication/SPI/SPI.cpp
```diff
@@ -31,30 +31,77 @@ uint8_t SPI::inscribe(SPI::Peripheral& spi) {
...
+
+    if (bind(spi_socket, (struct sockaddr*)&local_address,
+             sizeof(local_address)) < 0) {
+        ErrorHandler("Bind error: %s", strerror(errno));
```
g0nz4I0, line 62: nitpick, dont use ErrorHandler here

Src/HALALMock/Services/Communication/SPI/SPI.cpp
```diff
@@ -82,10 +129,21 @@ void SPI::assign_RS(uint8_t id, Pin& RSPin) {
...
-                          (GPIO_PinState)PinState::ON);
+    for (auto [_, spi] : SPI::registered_spi) {
+        SPI::init(spi);
+        EmulatedPin& SS_pin = SharedMemory::get_pin(*spi->SS);
+        SS_pin.PinData.SPI.is_on = true; // This pin tells the simulator to connect
+    }
+
+    // For each registered SPI master, accept the connection from the slave
+    for (auto& [id, sockets] : SPI::spi_master_sockets) {
+        auto& [master_socket, slave_socket] = sockets;
```
g0nz4I0, line 140: I like this

g0nz4I0 (changes requested):
Looks good, but if I remember correctly Runes.hpp had a setting specifying wether to use DMA or not, if DMA is specified you could try to use non-blocking send and recv operations.

And some general feedback, dont be afraid to terminate the program, this is not like the microcontroller which is harder to debug and gather info after unexpected results. just std::cout and terminate

jmaralo:
Fix the merge conflicts pls @jorgesg82

jmaralo (commented):
As is would work, but there are two things that can make the code better overall

Src/HALALMock/Services/Communication/SPI/SPI.cpp
```diff
@@ -189,27 +185,25 @@ bool SPI::transmit_and_receive(uint8_t id, span<uint8_t> command_data,
...
-                "%u",
-                (uint8_t)errorcode);
-            return false;
-            break;
-    }
+        if (recv(spi_master_sockets[id].second, receive_data.data(), receive_data.size(), MSG_WAITALL) < 0) {
+            ErrorHandler("Receive error: %s", strerror(errno));
+        }
+
+        HAL_SPI_TxRxCpltCallback(SPI::registered_spi[id]->hspi);
```
g0nz4I0, line 202: this has to go

Src/HALALMock/Services/Communication/SPI/SPI.cpp
```diff
@@ -160,26 +168,14 @@ bool SPI::receive(uint8_t id, span<uint8_t> data) {
...
-                "%u",
-                (uint8_t)errorcode);
-            return false;
-            break;
+    std::thread receiver_thread([id, data]() {
+        if (recv(spi_master_sockets[id].second, data.data(), data.size(), MSG_WAITALL) < 0) {
+        ErrorHandler("Receive error: %s", strerror(errno));
     }
+    });
+    receiverThread.detach();
```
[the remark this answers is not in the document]
g0nz4I0: I agree with this

Src/HALALMock/Services/Communication/SPI/SPI.cpp
```diff
@@ -189,27 +185,25 @@ bool SPI::transmit_and_receive(uint8_t id, span<uint8_t> command_data,
         return false;
     }
 
-    SPI::Instance* spi = SPI::registered_spi[id];
+    std::thread transmit_receive_thread([id, command_data, receive_data]() {
+        sockaddr_in broadcast_address;
+        broadcast_address.sin_family = AF_INET; // Set IPv4
+        broadcast_address.sin_port = htons(spi_master_sockets[id].second); // Set port
+        inet_pton(AF_INET, "192.168.0.255", &broadcast_address.sin_addr); // Set broadcast IP, supposing that whe have a mask = 255.255.255.0
```
g0nz4I0, line 192: I have some doubts about this

Inc/HALALMock/Services/Communication/SPI/SPI.hpp
```diff
@@ -346,6 +337,10 @@ class SPI{
...
      */
     static void init(SPI::Instance* spi);
+
+    static unordered_map<uint8_t, std::pair<int, int>> spi_master_sockets; /** Map that associates registered SPI master with its server socket and its port */
+
 };
 
-#endif
+extern std::array<int, 6> peripheral_ports; // Array of SPI peripheral ports, it should be defined on a .hpp file configuration
+extern std::string ip; // IP address of this VMCU, it should be defined on a .hpp file configuration
```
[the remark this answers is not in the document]
jorgesg82: If I put it inside SPI class, I can't use 'extern'

Src/HALALMock/Services/Communication/SPI/SPI.cpp
```diff
@@ -104,54 +139,27 @@ bool SPI::transmit(uint8_t id, span<uint8_t> data) {
...
+        }
+        return false;
     }
-}
 
-bool SPI::transmit_DMA(uint8_t id, span<uint8_t> data) {
-    if (!SPI::registered_spi.contains(id)) {
-        ErrorHandler("No SPI registered with id %u", id);
+    if (sendto(spi_master_sockets[id].first, data.data(), data.size(), 0, (struct sockaddr*)&broadcast_address, sizeof(broadcast_address)) < 0) {
+        ErrorHandler("Send error: %s", strerror(errno));
```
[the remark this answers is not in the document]
jorgesg82: I don't understand it. That's why I put data size as a parameter, isn't it? For the function to know that that's the amount of bytes it has to send. How it is possible that this function ends without send all those bytes?

Inc/HALALMock/Services/Communication/SPI/SPI.hpp
```diff
@@ -73,6 +73,22 @@ class SPI{
...
+
+        std::thread receiver_thread;
+        std::queue<span<uint8_t>> reception_queue;
+        std::mutex reception_mx;
+        std::condition_variable cv_reception;
+
+        std::thread sender_receiver_thread;
+        std::queue<std::pair<span<uint8_t>, span<uint8_t>>> transmission_reception_queue;
+        std::mutex transmission_reception_mx;
+        std::condition_variable cv_transmission_reception;
```
jmaralo, line 90: maybe use jthreads, they join when going out of scope

jmaralo (approved):
LGTM

Inc/HALALMock/Services/Communication/SPI/SPI.hpp
```diff
@@ -340,6 +356,8 @@ class SPI{
 
     static unordered_map<uint8_t, std::pair<int, int>> spi_master_sockets; /** Map that associates registered SPI master with its server socket and its port */
 
+    static void TxRxCpltCallback(SPI_HandleTypeDef* hspi);
+
```
g0nz4I0, line 360: the `SPI_HandleTypeDef` will make build fail

g0nz4I0 (approved):

PR #433, opened 2024-10-28
[FW-86] StateMachine state on SHM

jdmarmen:
1. Implemented a new shared memory for the state machine

- static void* emulated_state_machine is the pointer to which the shared memory will point when created
- uint8_t* state_machine_sm is the pointer that will be pointing to each byte of the emulated state machine after the cast of emulated_state_machine, as it will only be storing uint8_t's
- start_emulated_state_machine() creates the shared memory (size=16 bytes)
- update_state_machine_counter() updates the number of state machines that have been created, (that number is always stored in the first element of the memory array)
- update_current_state() changes de state id of a specific state machine

2. Added an implementation for incrementing the number of state machines and also storing their current states inside the shared memory

- Inside the StateMachine constructor, the state machine counter gets incremented when a new state machine is created, then, the initial state is stored on state_machine_sm[number of state machines] 
- Inside force_change_state(), before the actual state is about to be replaced by the new state, it gets stored in old_state, then, we revise each state that was already stored on the shared memory, in case we find the same id that was about to get replaced, we call update_current_state so as to rewrite the id with the newer one


Inc/HALALMock/Services/SharedMemory/SharedMemory.hpp
```diff
@@ -10,8 +10,15 @@ class SharedMemory {
 
     // do not use before SharedMemory::start, uninitialized!!!
     static EmulatedPin *gpio_memory;
+	static void* emulated_state_machine;
     static EmulatedPin &get_pin(Pin pin);
     static void start();
+	static void start_emulated_state_machine();
+	constexpr static size_t emulated_state_machine_size=16;
+	constexpr char* emulated_state_machine_name="/Shared_Memory_Emulated_State_Machine";
```
g0nz4I0, line 18: Don't hardcode the name, what happens if we launch two instances of the simulator?

g0nz4I0 (changes requested):
I believe all the changes inside SharedMemory.hpp and SharedMemory.cpp should actually be in StateMachine.hpp and StateMachine.cpp. And be called on the constructor of StateMachine

Inc/HALALMock/Services/SharedMemory/SharedMemory.hpp
jdmarmen: Why would we launch to instances of the simulator, this shared memory should only be used for the state machine right? If not, should we generate an individual name for each time we launch an instance??

g0nz4I0: two instances would be used for example if we wanted to run two boards in the same computer

Src/HALALMock/Services/SharedMemory/SharedMemory.cpp
```diff
@@ -1,9 +1,60 @@
...
+#include <sys/stat.h>
+#include <sys/mman.h>
+#include <unistd.h>
 
 void SharedMemory::start() {
     // TODO: create / open shared memory
 }
+
+std::string SharedMemory::generate_shared_memory_name(){
+	return "SharedMemory_emulated_state_machine_"+std::to_string(getpid());
```
g0nz4I0, line 15: this wont work if another process wants to connect to it, because it wont know the PID

Inc/HALALMock/Services/SharedMemory/SharedMemory.hpp
```diff
@@ -10,8 +10,16 @@ class SharedMemory {
...
+	static void* emulated_state_machine{};
     static EmulatedPin &get_pin(Pin pin);
     static void start();
+	static void start_emulated_state_machine();
+	constexpr static size_t emulated_state_machine_size=16;
+	static std::string generate_shared_memory_name();
+	static std::string emulated_state_machine_name{};
+	static uint8_t* state_machine_sm{};
+	static void update_state_machine_counter(uint8_t counter);
+	static void update_current_state(uint8_t index, uint8_t state);
```
g0nz4I0, line 22: I believe all the changes inside SharedMemory.hpp and SharedMemory.cpp should actually be in StateMachine.hpp and StateMachine.cpp. And be called on the constructor of StateMachine

Src/HALALMock/Services/SharedMemory/SharedMemory.cpp
```diff
@@ -1,9 +1,60 @@
...
+	if(emulated_state_machine==MAP_FAILED){
+		std::cout<<"Error mapping the shared memory object\n";
+		close(shm_state_machine_fd);
+		std::terminate();
+	}
+
+	// cast of the void pointer that the shared memory object returns to a uint8_t pointer
+	state_machine_sm=static_cast<uint8_t*>(emulated_state_machine);
+	// initialize the shared memory counter by setting it to 0, when a new state machine is created it will increment
+	state_machine_sm[0]=0;
```
g0nz4I0, line 46: oh, this will crash your program, because when it gets initialized it gets the value `nullptr`, which you cannot derefence, if you want to keep this logic just make a call to `malloc()` specifying some buffer so that you are sure it is your memory, ideally use unique_ptr so you free the memory whenever the program ends.

g0nz4I0 (changes requested):
What I said in the last review, move the state machine related stuff from SharedMemory to StateMachine, the logic looks good

jmaralo:
@g0nz4I0 For the SharedMemory initialization I think it makes sense to create the SHM in the class named SharedMemory, my idea of it is that all the configuration related to the shared memory (independent of weather it is for GPIO pins or the state machine) is in one place. I see your point that it could be alongside the StateMachine, but I feel like the SharedMemory class is a better place for it as you just need to change one file to update any shared memory configuration, and the whole class is dedicated to just that

Inc/HALALMock/Services/SharedMemory/SharedMemory.hpp
```diff
@@ -10,8 +10,16 @@ class SharedMemory {
 
     // do not use before SharedMemory::start, uninitialized!!!
     static EmulatedPin *gpio_memory;
+	static void* emulated_state_machine{};
```
jmaralo, line 13: you don't need to keep this variable around, just static_cast the result from mmap inside the function

Inc/HALALMock/Services/SharedMemory/SharedMemory.hpp
jmaralo: (for @g0nz4I0) Notice that the shared memory is for all state machines, not just one, so having its creation, manipulation and destruction tied to those of the state machine won't work as the shared memory will probably live even longer than the program itself. Having it here makes more sense to me as this is the class for all Shared Memory configuration and manipulation, because in that case we would have moved all the GPIO SHM stuff to `PinModel.hpp` and `PinModel.cpp` instead of creating a dedicated class.

Src/HALALMock/Services/SharedMemory/SharedMemory.cpp
```diff
@@ -1,9 +1,60 @@
...
+#include <sys/mman.h>
+#include <unistd.h>
 
 void SharedMemory::start() {
     // TODO: create / open shared memory
 }
+
+std::string SharedMemory::generate_shared_memory_name(){
+	return "SharedMemory_emulated_state_machine_"+std::to_string(getpid());
+}
```
jmaralo, line 16: use extern variable as I said, if you don't know well what an exern is take a look at [this](https://www.learncpp.com/cpp-tutorial/external-linkage-and-variable-forward-declarations/) or any other resources you find online.

Src/HALALMock/Services/SharedMemory/SharedMemory.cpp
jmaralo: (for @g0nz4I0) You are mixing things, this pointer comes from the mmap call, it has just been casted to `uint8_t*`, so it will be valid, no need to call malloc because all that memory allocation has already been performed by shm_open, ftruncate and mmap

Src/ST-LIB_LOW/StateMachine/StateMachine.cpp
```diff
@@ -105,6 +109,13 @@ StateMachine::StateMachine(uint8_t initial_state) :
 	initial_state(initial_state), current_state(initial_state) {
 	add_state(initial_state);
 	enter_state(initial_state);
+
+	#ifdef SIM_ON
+		SharedMemory::start_emulated_state_machine();
```
jmaralo, line 114: this should be started only on SharedMemory::start(), which will in turn be called on HALAL::start()

Src/ST-LIB_LOW/StateMachine/StateMachine.cpp
```diff
@@ -105,6 +109,13 @@ StateMachine::StateMachine(uint8_t initial_state) :
...
+		uint8_t counter=SharedMemory::state_machine_sm[0]+1;
```
jmaralo, line 115:
```suggestion
		uint8_t counter = SharedMemory::state_machine_sm[0]++;
```

jmaralo, line 115: store this counter as part of the state machine, later you are searching for the old state, but multiple state machines might have the same ID for a state, which won't work with your logic

Src/ST-LIB_LOW/StateMachine/StateMachine.cpp
```diff
@@ -269,11 +280,20 @@ void StateMachine::force_change_state(uint8_t new_state) {
...
 	current_state = new_state;
 
 	enter_state(current_state);
 	register_all_timed_actions(current_state);
+
+	#ifdef SIM_ON
+		for(uint8_t i=1; i<=SharedMemory::state_machine_sm[0]; i++){
+			if(SharedMemory::state_machine_sm[i] == old_state)
+				SharedMemory::update_current_state(i,current_state);
+		}
```
jmaralo, line 295:
```suggestion
        SharedMemory::update_current_state(the_counter_you_stored_in_the_constructor, current_state);
```
Simplifies a lot the logic

Src/ST-LIB_LOW/StateMachine/StateMachine.cpp
```diff
@@ -269,11 +280,20 @@ void StateMachine::force_change_state(uint8_t new_state) {
 
 	unregister_all_timed_actions(current_state);
 	exit_state(current_state);
-
+	#ifdef SIM_ON
+		old_state=current_state;
+	#endif
```
jmaralo, line 285: no need to store the previos state to search then for it, look at the other comments

jmaralo (changes requested):
Main thing is that when you create the state machine, the current SHM counter is used as kind of an "ID" for that state machine which you can then use to change the stored state, no need to search for the value that changed or anything like that, which simplifies a lot the logic

Src/ST-LIB_LOW/StateMachine/StateMachine.cpp
jdmarmen: What do you mean by storing the name as part of the state machine, creating a public instance variable at StateMachine.hpp that points to that value???

Inc/HALALMock/Services/SharedMemory/SharedMemory.hpp
```diff
@@ -10,8 +10,16 @@ class SharedMemory {
...
     static EmulatedPin &get_pin(Pin pin);
     static void start();
+	static void start_emulated_state_machine();
+	constexpr static size_t emulated_state_machine_size=16;
+	static std::string generate_shared_memory_name();
+	static std::string emulated_state_machine_name{};
```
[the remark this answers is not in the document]
jdmarmen: ok, i'll store them inside the start function. also thought on doing the PID thing but storing each name inside a file, for other processes to consult

Src/HALALMock/Services/SharedMemory/SharedMemory.cpp
g0nz4I0: xd, I dont know what  I read, honestly I though the pointer wasn't even initialized, my bad.

Inc/HALALMock/Services/SharedMemory/SharedMemory.hpp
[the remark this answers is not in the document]
g0nz4I0: we need to find a way of naming the shared memories so we are consistent

Src/ST-LIB_LOW/StateMachine/StateMachine.cpp
g0nz4I0: probably just get rid of the `update_state_machine_counter()`, and simply do `SharedMemory::state_machine_sm[0]++`  because this suggestion you just did does exactly the opposite of what you want it to do, take a look.

Src/ST-LIB_LOW/StateMachine/StateMachine.cpp
g0nz4I0: yes, this is it.

Inc/HALALMock/Services/SharedMemory/SharedMemory.hpp
g0nz4I0: absolutely

g0nz4I0 (changes requested):
Overall  implement the change of storing the id of the state machine when creating it, it is eaiser  to index into the shared_memory this way

Src/ST-LIB_LOW/StateMachine/StateMachine.cpp
jmaralo: to create a new private member of the StateMachine class named something like `shm_id` where you store the counter and then you use that variable to access the SHM

g0nz4I0: noup, in any case you should pre-increment

g0nz4I0: no, I guess more like this:
```c++
uint8_t state_machine_id_in_shm;
static uint8_t state_machines_count{0};
```
In constructor:
```c++
state_machine_id_in_shm = ++state_machines_count;
```
and in update_sm(), this cannot be static anymore if you do something like what I mentioned
```c++
state_machine_sm[state_machine_id_in_shm] = current_state;
```

Src/ST-LIB_LOW/StateMachine/StateMachine.cpp
```diff
@@ -105,6 +109,11 @@ StateMachine::StateMachine(uint8_t initial_state) :
 	initial_state(initial_state), current_state(initial_state) {
 	add_state(initial_state);
 	enter_state(initial_state);
+
+	#ifdef SIM_ON
+		state_machine_id_in_shm = SharedMemory::state_machine_memory[0]++;
+		SharedMemory::update_current_state(state_machine_id_in_shm,initial_state);
+	#endif
```
g0nz4I0, line 116: this does not work, lets think of an example:

First execution:
```c++
state_machine_memory[0] = 0 // this was done in shared_memory_start
//now:
// because you are using postfix increment 
// check this example https://godbolt.org/z/Gn6ef4T64
state_machine_id_in_shm = 0; 
// now you call update_current_state
// 0 is your state machine id
// if initial state is 0 for example you have a loop
//if it is different than one, for example, 6 what happens,

state_machine_memory[0] = initial_state

// on the next iteration you have that state_machine_memory[0] = 6
// and so the next id that you will assign will be 6

```

g0nz4I0 (changes requested):

Inc/HALALMock/Services/SharedMemory/SharedMemory.hpp
```diff
@@ -12,6 +12,11 @@ class SharedMemory {
     static EmulatedPin *gpio_memory;
     static EmulatedPin &get_pin(Pin pin);
     static void start();
+	static uint8_t* state_machine_memory{};
+	static void start_state_machine_memory();
+	constexpr static size_t state_machine_memory_size=16;
+	extern constexpr char* state_machine_memory_name;
+	static void update_current_state(uint8_t index, uint8_t state);
```
jmaralo, line 19:
```suggestion
    static EmulatedPin *gpio_memory;
	static uint8_t* state_machine_memory{};
	
	constexpr static size_t state_machine_memory_size=16;
	extern const char* state_machine_memory_name;
	
    static void start();
    static void start_state_machine_memory();
    
    static EmulatedPin &get_pin(Pin pin);
    
    static void update_current_state(uint8_t index, uint8_t state);
```

Inc/HALALMock/Services/SharedMemory/SharedMemory.hpp
```diff
@@ -12,6 +12,11 @@ class SharedMemory {
...
 };
```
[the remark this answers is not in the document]
jmaralo: this is stored on the first byte of the SHM so that the simulator can know from the outside how many state machines are initialized

Src/ST-LIB_LOW/StateMachine/StateMachine.cpp
```diff
@@ -105,6 +109,11 @@ StateMachine::StateMachine(uint8_t initial_state) :
 	initial_state(initial_state), current_state(initial_state) {
 	add_state(initial_state);
 	enter_state(initial_state);
+
+	#ifdef SIM_ON
+		state_machine_id_in_shm = SharedMemory::state_machine_memory[0]++;
```
jmaralo, line 114: as per gonzalo said
```suggestion
		state_machine_id_in_shm = ++SharedMemory::state_machine_memory[0];
```

Src/ST-LIB_LOW/StateMachine/StateMachine.cpp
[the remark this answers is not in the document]
jmaralo: the idea is for the first byte of the SHM to act as a counter so the python knows how many state machines there are

jmaralo (changes requested):
Mostly what gonzalo said and just a refactor to the SharedMemory class, which had the definitions in the header file all around the place

jmaralo (approved):

g0nz4I0 (approved):

PR #438, opened 2024-10-30
[FW-79]- Creation of ethernet Mock

g0nz4I0:


Inc/HALALMock/Models/IPV4/IPV4.hpp
```diff
@@ -1,26 +1,23 @@
...
-
+#include <atomic> 
+#include <mutex> 
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <netinet/tcp.h> 
+#include <unistd.h>
+#include <arpa/inet.h>
+#include <thread>
 #ifdef HAL_ETH_MODULE_ENABLED
```
g0nz4I0, line 11: remove this line

Inc/HALALMock/Services/Communication/Ethernet/Ethernet.hpp
```diff
@@ -10,13 +10,10 @@
 #include "C++Utilities/CppUtils.hpp"
 #include "HALALMock/Models/PinModel/Pin.hpp"
 #include "HALALMock/Services/Communication/Ethernet/EthernetNode.hpp"
-#include "lwip.h"
-#include "ethernetif.h"
+//#include "lwip.h"
+//#include "ethernetif.h"
```
g0nz4I0, line 14: dont comment, remove

Inc/HALALMock/Services/Communication/Ethernet/Ethernet.hpp
```diff
@@ -10,13 +10,10 @@
...
 #include "HALALMock/Services/Communication/Ethernet/EthernetHelper.hpp"
 #ifdef HAL_ETH_MODULE_ENABLED
```
g0nz4I0, line 16: remove

Inc/HALALMock/Services/Communication/Ethernet/TCP/ServerSocket.hpp
```diff
@@ -43,14 +43,14 @@ class ServerSocket : public OrderProtocol{
 
 	static unordered_map<uint32_t,ServerSocket*> listening_sockets;
 	struct tcp_pcb* server_control_block = nullptr;
-	queue<struct pbuf*> tx_packet_buffer;
-	queue<struct pbuf*> rx_packet_buffer;
+	//queue<struct pbuf*> tx_packet_buffer;
+	//queue<struct pbuf*> rx_packet_buffer;
```
g0nz4I0, line 47: probably keep but change inner type, keep the std::queue

Inc/HALALMock/Services/Communication/Ethernet/TCP/ServerSocket.hpp
```diff
@@ -43,14 +43,14 @@ class ServerSocket : public OrderProtocol{
...
 	IPV4 local_ip;
 	uint32_t local_port;
 	IPV4 remote_ip;
 	ServerState state;
 	static uint8_t priority;
-	struct tcp_pcb* client_control_block;
+	//struct tcp_pcb* client_control_block;
```
g0nz4I0, line 53: remove, dont comment

Inc/HALALMock/Services/Communication/Ethernet/TCP/Socket.hpp
```diff
@@ -11,11 +11,20 @@
 #include "HALALMock/Models/Packets/Packet.hpp"
 #include "HALALMock/Models/Packets/Order.hpp"
 #include "HALALMock/Models/Packets/OrderProtocol.hpp"
+#include <cstring>
+#include <iostream>
 #ifdef HAL_ETH_MODULE_ENABLED
```
g0nz4I0, line 16: remove

Inc/HALALMock/Services/Communication/Ethernet/TCP/Socket.hpp
```diff
@@ -27,11 +36,12 @@ class Socket : public OrderProtocol{
...
 	IPV4 remote_ip;
 	uint32_t remote_port;
-	tcp_pcb* connection_control_block;
-	tcp_pcb* socket_control_block;
 	SocketState state;
-	queue<struct pbuf*> tx_packet_buffer;
-	queue<struct pbuf*> rx_packet_buffer;
+	//Change the data type of tx/rx packet_buffer
+	queue<vector<uint8_t>> tx_packet_buffer;
+	queue<vector<uint8_t>> rx_packet_buffer;
```
g0nz4I0, line 42: probably just a vector of `Packet*` not of `vector<uint8_t>`

Src/HALALMock/Services/Communication/Ethernet/Ethernet.cpp
```diff
@@ -6,66 +6,17 @@
...
-	  MPU_InitStruct.Number = MPU_REGION_NUMBER3;
-	  MPU_InitStruct.BaseAddress = 0x08000000;
-	  MPU_InitStruct.Size = MPU_REGION_SIZE_1MB;
-	  MPU_InitStruct.DisableExec = MPU_INSTRUCTION_ACCESS_ENABLE;
-	  MPU_InitStruct.IsBufferable = MPU_ACCESS_NOT_BUFFERABLE;
-	  HAL_MPU_ConfigRegion(&MPU_InitStruct);
-	  /* Enables the MPU */
-	  HAL_MPU_Enable(MPU_PRIVILEGED_DEFAULT);
-}
+void Ethernet::mpu_start(){}
```
g0nz4I0, line 19: good

Src/HALALMock/Services/Communication/Ethernet/Ethernet.cpp
```diff
@@ -88,57 +39,41 @@ void Ethernet::start(IPV4 local_ip, IPV4 subnet_mask, IPV4 gateway){
...
 	}
-
-
 }
-
 void Ethernet::update(){
+	//I'm going to leave the case is not running so it warn the user to check if you have done inscribed
 	if(not is_running) {
-		ErrorHandler("Ethernet is not running, check if its been inscribed");
+		std::cout<<"Ethernet is not running, check if its been inscribed"<<std::endl;
```
g0nz4I0, line 74: ErrorHandler here, its a board problem

Src/HALALMock/Services/Communication/Ethernet/TCP/Socket.cpp
```diff
@@ -101,185 +177,104 @@ void Socket::reset(){
...
+}
 void Socket::process_data(){
+	
 	while(!rx_packet_buffer.empty()){
-		struct pbuf* packet = rx_packet_buffer.front();
-		rx_packet_buffer.pop();
-		uint8_t* new_data = (uint8_t*)(packet->payload);
-		tcp_recved(socket_control_block, packet->tot_len);
+		{
+			std::std::lock_guard<std::mutex> lock(mtx);
```
g0nz4I0, line 234: std::std:: typo I guess

Src/HALALMock/Services/Communication/Ethernet/TCP/Socket.cpp
```diff
@@ -101,185 +177,104 @@ void Socket::reset(){
...
+    std::vector<uint8_t> order_buffer = order.build();
+    if (order_buffer.empty()) {
+        std::cout << "Error: building de order buffer\n";
+        return false; 
+    }
+    std::vector<uint8_t> packet(order_buffer); 
+    {
+        std::lock_guard<std::mutex> lock(mutex); 
+        tx_packet_buffer.push(packet); 
+    }
```
g0nz4I0, line 256: good job

Src/HALALMock/Services/Communication/Ethernet/TCP/Socket.cpp
```diff
@@ -101,185 +177,104 @@ void Socket::reset(){
...
 
 bool Socket::is_connected(){
 	return state == Socket::SocketState::CONNECTED;
 }
 
-err_t Socket::connect_callback(void* arg, struct tcp_pcb* client_control_block, err_t error){
-	IPV4 remote_ip;
-	remote_ip.address = client_control_block->remote_ip;
-	EthernetNode remote_node(remote_ip,client_control_block->remote_port);
+err_t Socket::connect_callback(void* arg, struct tcp_pcb* client_control_block, err_t error){}
```
g0nz4I0, line 264: noup, you have to delete the callbacks, or change the interface, there is a param that is `struct_tcp_pcb*` that is an LwIp data type

Src/HALALMock/Services/Communication/Ethernet/TCP/Socket.cpp
```diff
@@ -68,31 +137,38 @@ Socket::Socket(IPV4 local_ip, uint32_t local_port, IPV4 remote_ip, uint32_t remo
...
+	timeout.tv_usec = 0;
+	//use select to check if fthe socket has established connection
+	int result = select(socket_fd + 1, NULL, &write_fds, NULL, &timeout);
+	if (result > 0 && FD_ISSET(socket_fd, &write_fds)) {
+		//If it's connected we do as connected_callback
+		if(connecting_sockets.contains(remote_node)){
+			connecting_sockets.erase(remote_node);
+			state = CONNECTED;
+		}
+	}
```
g0nz4I0, line 171: this timeout I dont understand

Src/HALALMock/Services/Communication/Ethernet/TCP/Socket.cpp
```diff
@@ -37,28 +30,104 @@ Socket::~Socket(){
...
```
g0nz4I0, line 129: this timeout I dont understand

Src/HALALMock/Services/Communication/Ethernet/Ethernet.cpp
```diff
@@ -88,57 +39,41 @@ void Ethernet::start(IPV4 local_ip, IPV4 subnet_mask, IPV4 gateway){
...
+		Pin pin_list_ethernet[number_pin_ethernet] = {PA1,PA2,PA7,PB13,PC1,PC4,PC5,PG11,PG0,PG13};
+		for(size_t i = 0; i < number_pin_ethernet; i++){
+			EmulatedPin &pin_data = SharedMemory::get_pin(pin_list_ethernet[i]);
+			if(pin_data.type == PinType::NOT_USED){
+			pin_data.type = PinType::Ethernet;
+			}else{
+				std::cout<<"Error inscribing ethernet Pins, PA1,PA2,PA7,PB13,PC1,PC4,PC5,PG11,PG0,PG13 must be free"<<std::endl;
+				return;
+			}
+		}
```
g0nz4I0, line 66: I'm not sure if we have to do this, but it doesn't hurt

g0nz4I0 (commented):
Overall looks good, I dont understand the purpose of the timeout, and if you can install callbacks maybe we can do that? instead of the thread. IPV4 interface I like the small changes

Remove all the #ifdef guards 
I like the `std::lock_guard`

Inc/HALALMock/Services/Communication/Ethernet/TCP/ServerSocket.hpp
oganigl: yes I haven't start with serverSocket yet

Inc/HALALMock/Models/IPV4/IPV4.hpp
oganigl: I haven't seen in any place a real callback to do once the connection is done thats why I used the timeout to wait and if it's connected do the callback routine.

Src/HALALMock/Services/Communication/Ethernet/TCP/Socket.cpp
oganigl: Okay I didn't want it to change it yet due to everything is public, I will change it to private in the HALAL and change the interface the problem, will be the one to set the keepalives

Src/HALALMock/Services/Communication/Ethernet/Ethernet.cpp
oganigl: I think if some pins are being used in ethernet we should, have them in the gpio_memory so if someone try to access for a pwm they receive an error message

Inc/HALALMock/Services/Communication/Ethernet/TCP/Socket.hpp
oganigl: Do you thing is necesarry to change it due to I think now the implementacion with vetor<uitnt8> could work.

Src/HALALMock/Services/Communication/Ethernet/TCP/Socket.cpp
```diff
@@ -101,185 +177,104 @@ void Socket::reset(){
...
+    while (is_receiving) {
+        uint8_t buffer[1024]; // Buffer for the data
+        ssize_t received_bytes = ::recv(socket_fd, buffer, sizeof(buffer), 0);
+        if (received_bytes > 0) {
+            std::vector<uint8_t> packet(buffer, buffer + received_bytes); 
+            {
+                std::lock_guard<std::mutex> lock(mtx); 
+                rx_packet_buffer.push(packet);
+				process_data(); 
+            }
```
jmaralo, line 221: Fun fact, TCP works in streams and we are using nonblocking mode, so the previous call to received will return all the packets that have arrived and can even return half of a packet in one call and the other half in the next. I don't know how the library was designed, but if someone made the assumption that the data passed to the packet constructor or Order::process_data was just one full packet (like UDP does it) this won't work and it will be a pain in the ass to fix, just letting everyone know.

jmaralo: not really related to this PR, but it's important to know

jmaralo (changes requested):
There are some corner cases that I pointed out that I got after working two years with sockets from Go which might get you if you don't know them.

I also thing there is a lot of refactoring that can be done to make the code much cleaner and easy to follow. Is easy for the code to get messy with all the syscalls and error handling ifs.

Src/HALALMock/Services/Communication/Ethernet/TCP/Socket.cpp
g0nz4I0: the packets need to match EXACTLY, otherwise bad things will happen

g0nz4I0: we might have to change something in the impl actually

Src/HALALMock/Services/Communication/Ethernet/TCP/Socket.cpp
```diff
@@ -101,185 +177,104 @@ void Socket::reset(){
...
+    if (is_receiving) {
+        is_receiving = false;
+        if (receiving_thread.joinable()) {
+            receiving_thread.join(); // wait until the thread finish
+        }
+    }
+}
+void Socket::receive() {
+    while (is_receiving) {
+        uint8_t buffer[1024]; // Buffer for the data
```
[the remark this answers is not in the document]
g0nz4I0: define should befine, extern is for things the user might want to configure, this is a developer setting.

Src/HALALMock/Services/Communication/Ethernet/TCP/Socket.cpp
```diff
@@ -101,185 +177,104 @@ void Socket::reset(){
...
+        ssize_t received_bytes = ::recv(socket_fd, buffer, sizeof(buffer), 0);
```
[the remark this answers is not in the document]
g0nz4I0: isn't this in a thread? 
make this a jthread, so when the program terminates it gets called in the destructor

Src/HALALMock/Services/Communication/Ethernet/TCP/Socket.cpp
```diff
@@ -101,185 +177,104 @@ void Socket::reset(){
...
+                std::lock_guard<std::mutex> lock(mtx); 
+                rx_packet_buffer.push(packet);
+				process_data(); 
+            }
+        } else if (received_bytes < 0) {
+            std::cout << "Error receiving data\n";
+			socket->state = CLOSING;
+            stop_receiving(); 
+			close();
+        }
```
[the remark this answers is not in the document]
g0nz4I0: this wont go well indeed

Src/HALALMock/Services/Communication/Ethernet/TCP/Socket.cpp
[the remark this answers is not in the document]
g0nz4I0: yes

Src/HALALMock/Services/Communication/Ethernet/TCP/Socket.cpp
```diff
@@ -37,28 +30,104 @@ Socket::~Socket(){
...
+        return;
+    }
+	 // Configure TCP_KEEPCNT (number keepalives are send before considering the connection down)
+	 float keep_cnt = static_cast<float>(keepalive_config.tries_until_disconnection)/1000.0;
+    if (setsockopt(socket_fd, IPPROTO_TCP, TCP_KEEPCNT, &keep_cnt, sizeof(keep_cnt)) < 0) {
+        std::cout << "Error to configure TCP_KEEPCNT\n";
+        close(socket_fd);
+        return ;
+    }
+	connecting_sockets[remote_node] = this;
```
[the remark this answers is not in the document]
g0nz4I0: Noooo, clean code merchant :(

Src/HALALMock/Services/Communication/Ethernet/TCP/Socket.cpp
```diff
@@ -37,28 +30,104 @@ Socket::~Socket(){
...
-	connecting_sockets[remote_node] = this;
-	tcp_connect(connection_control_block, &remote_ip.address , remote_port, connect_callback);
+	tcp_connection_sim();
 	OrderProtocol::sockets.push_back(this);
 }
+Socket::tcp_connection_sim(){
+	//create socket
+	socket_fd = ::socket(AF_INET,SOCK_STREAM,0);
+	//create socket non-blocking
+	int status = fcntl(socketfd, F_SETFL, fcntl(socketfd, F_GETFL, 0) | O_NONBLOCK);
```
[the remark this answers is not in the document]
oganigl: sockets
I'm not sure if you can see the image so I put the link:https://stackoverflow.com/questions/1543466/how-do-i-change-a-tcp-socket-to-be-non-blocking
and the message:
To mark a socket as non-blocking the code is as simple as:

// where socketfd is the socket you want to make non-blocking
int status = fcntl(socketfd, F_SETFL, fcntl(socketfd, F_GETFL, 0) | O_NONBLOCK);

if (status == -1){
  perror("calling fcntl");
  // handle the error.  By the way, I've never seen fcntl fail in this way
}
Under Linux, on kernels > 2.6.27 you can also create sockets non-blocking from the outset using [socket()](http://man7.org/linux/man-pages/man2/socket.2.html#DESCRIPTION) and [accept4()](https://linux.die.net/man/2/accept4).

e.g.

   // client side
   int socketfd = socket(AF_INET, SOCK_STREAM | SOCK_NONBLOCK, 0);

   // server side - see man page for accept4 under linux 
   int socketfd = accept4( ... , SOCK_NONBLOCK);
It saves a little bit of work, but is less portable so I tend to set it with fcntl().

Summing up: it says that for old linux you need to do what I did last linux let you do what you said. To make it able to use in any linux that`s why I decided used the first one. What do you think?

Src/HALALMock/Services/Communication/Ethernet/TCP/Socket.cpp
```diff
@@ -101,185 +177,104 @@ void Socket::reset(){
...
+void Socket::start_receiving(){
+	is_receiving = true;
+    receiving_thread = std::thread(&Socket::receive, this);
+}
+void Socket::stop_receiving() {
+    if (is_receiving) {
+        is_receiving = false;
+        if (receiving_thread.joinable()) {
+            receiving_thread.join(); // wait until the thread finish
+        }
```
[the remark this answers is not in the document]
oganigl: Yes, Juan I check it and you're right. The problem is that I haven't find any function to terminate the thread without killing the whole process

Src/HALALMock/Services/Communication/Ethernet/TCP/Socket.cpp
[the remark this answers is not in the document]
oganigl: what happens if I call reset, there is no way to terminate the thread to create a new one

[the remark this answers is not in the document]
oganigl: ???

Inc/HALALMock/Services/Communication/Ethernet/TCP/Socket.hpp
```diff
@@ -37,9 +41,8 @@ class Socket : public OrderProtocol{
 	IPV4 remote_ip;
 	uint32_t remote_port;
 	SocketState state;
-	//Change the data type of tx/rx packet_buffer
-	queue<vector<uint8_t>> tx_packet_buffer;
-	queue<vector<uint8_t>> rx_packet_buffer;
+	queue<Packet> tx_packet_buffer;
+	queue<Packet> rx_packet_buffer;
```
g0nz4I0, line 45: this should be a reference or a pointer, otherwise you dont get polymorphism

Src/HALALMock/Services/Communication/Ethernet/TCP/Socket.cpp
```diff
@@ -138,6 +131,13 @@ Socket::Socket(IPV4 local_ip, uint32_t local_port, IPV4 remote_ip, uint32_t remo
 Socket::Socket(EthernetNode local_node, EthernetNode remote_node):Socket(local_node.ip, local_node.port, remote_node.ip, remote_node.port){}
 
 void Socket::close(){
+	if(is_connecting){
+		is_connecting = false;
+		connection_thread;
```
g0nz4I0, line 136: what is this trying to do?

Src/HALALMock/Services/Communication/Ethernet/TCP/Socket.cpp
```diff
@@ -180,16 +168,14 @@ void Socket::reset(){
...
-
+	configure_socket_and_connect();
 }
 
 
 void Socket::send(){
 	while (!tx_packet_buffer.empty()) {
-        vector<uint8_t> packet = tx_packet_buffer.front();
-        ssize_t sent_bytes = ::send(socket_fd, packet.data(), packet.size(), 0);
+        HeapPacket packet = tx_packet_buffer.front();
```
g0nz4I0, line 177: Pointer or reference type, Packet* will do

g0nz4I0:
⚠️ Cuando tienes tipos polimórficos tienes que utilizarlos como pointer types, esto es, como una referencia o como un puntero, si no no te va a funcionar, Packet en este caso es la clase base de todo, es decir, todos heredan de packet.⚠️

Inc/HALALMock/Services/Communication/Ethernet/TCP/Socket.hpp
oganigl: True

Src/HALALMock/Services/Communication/Ethernet/TCP/Socket.cpp
oganigl: now nothing this is the things I didn't nknow how to solve it I'm going to put a destructor

Src/HALALMock/Services/Communication/Ethernet/TCP/Socket.cpp
```diff
@@ -133,19 +142,21 @@ Socket::Socket(EthernetNode local_node, EthernetNode remote_node):Socket(local_n
 void Socket::close(){
 	if(is_connecting){
 		is_connecting = false;
-		connection_thread;
+		~connection_thread;
```
g0nz4I0, line 145: `~connection_thread()`

Src/HALALMock/Services/Communication/Ethernet/TCP/Socket.cpp
```diff
@@ -174,8 +198,8 @@ void Socket::reset(){
 
 void Socket::send(){
 	while (!tx_packet_buffer.empty()) {
-        HeapPacket packet = tx_packet_buffer.front();
-        ssize_t sent_bytes = ::send(socket_fd, packet.build(), packet.get_size(), 0);
+        HeapPacket *packet = tx_packet_buffer.front();
```
g0nz4I0, line 201: Packet*

oganigl: can you explain this, at the end packet is a virtual class to store data shouldn't I use one class derivated from packet

Src/HALALMock/Services/Communication/Ethernet/TCP/ServerSocket.cpp
```diff
@@ -163,7 +141,128 @@ void ServerSocket::send(){
...
+        close(server_socket_fd);
+        return;
+    }
+	 // Configure TCP_KEEPCNT (number keepalives are send before considering the connection down)
+	 float keep_cnt = static_cast<float>(keepalive_config.tries_until_disconnection)/1000.0;
+    if (setsockopt(server_socket_fd, IPPROTO_TCP, TCP_KEEPCNT, &keep_cnt, sizeof(keep_cnt)) < 0) {
+        std::cout << "Error to configure TCP_KEEPCNT\n";
+        close(server_socket_fd);
+        return ;
+    }
```
jmaralo, line 203: I would still separate into more functions, but just a nitpick, also, return a `bool` that is `true` if the configuration was successful or `false` in any other case

Src/HALALMock/Services/Communication/Ethernet/UDP/DatagramSocket.cpp
```diff
@@ -43,9 +24,44 @@ DatagramSocket::~DatagramSocket(){
...
+		while(true){
+			uint8_t received_data[1024];
+			struct sockaddr_in src_addr;
+			socklen_t addr_len = sizeof(src_addr);
+			ssize_t size = recvfrom(udp_socket,(uint8_t*)received_data,MAX_SIZE_PACKET,0,(struct sockaddr *)&src_addr, &addr_len);
+			if(size < 0){
+				cout<<"Error in function recvfrom\n";
+				is_receiving = false;
+				return;
+			}
```
jmaralo, line 55: don't know if size can be negative, what is `ssize_t`?

jmaralo (commented):
Gave it a quick read and didn't find anything weird, keep going at it so we can start seeing where compilation fails. But I'll point out just two things

Src/HALALMock/Services/Communication/Ethernet/UDP/DatagramSocket.cpp
oganigl: image

oganigl:
I don't know how to wite a description of the pr once I changed it from draft to open, so I will leave this message explaining the pr. During this pr I've tried to convert the code from the lwip to sockets in linux. But always trying to replicate the behave of the halal. we will have to create test to assure the well behave of the pr. The changes have been done in these files: Ethernet, Pin,IPV4,  Socket, ServerSocket and DatagramSocket

Src/HALALMock/Services/Communication/Ethernet/TCP/ServerSocket.cpp
```diff
@@ -1,43 +1,20 @@
...
 
 uint8_t ServerSocket::priority = 1;
 unordered_map<uint32_t,ServerSocket*> ServerSocket::listening_sockets = {};
-
 ServerSocket::ServerSocket() = default;
 
 ServerSocket::ServerSocket(IPV4 local_ip, uint32_t local_port) : local_ip(local_ip),local_port(local_port){
 	if(not Ethernet::is_running) {
-		ErrorHandler("Cannot declare UDP socket before Ethernet::start()");
+		cout<<"Cannot declare UDP socket before Ethernet::start()\n";
```
jmaralo, line 11: ErrorHandler

Src/HALALMock/Services/Communication/Ethernet/TCP/ServerSocket.cpp
```diff
@@ -72,209 +47,230 @@ ServerSocket::~ServerSocket(){
...
+        tx_packet_buffer.pop();
+    }
+    while (!rx_packet_buffer.empty()) {
+        rx_packet_buffer.pop();
+    }
+	//eliminate the threads
+	if(state == LISTENING){
+		~listening_thread();
+	}else if(state == ACCEPTED){
+		~receive_thread();
```
jmaralo, line 70: I dont know if closing both in the `close` method and in the destructor will give any problems, because usually doing a double close returns an error

jmaralo: also see that in the HALAL the implementation is different for close and for the destructor

jmaralo (approved):
I'm approving the PR because it has been open for too long. There are a few things that won't work with ServerSocket from the way it worked with LWIP and the way it works in linux. I think we should merge this PR and fix those things later on with new jira tickets

oganigl (approved):
I think this pr it has to been check by testing it.

PR #439, opened 2024-10-30
[Fw-80] HalalMock FDCAN 

Cantonplas:
Added support for FDCAN communications for HalalMock based on sockets and UDP packets to simulate FDCAN.


Src/HALALMock/Services/Communication/FDCAN/FDCAN.cpp
```diff
@@ -89,43 +92,52 @@ bool FDCAN::transmit(uint8_t id, uint32_t message_id, const char* data, FDCAN::D
...
+	if(recv(instance->socket, data->rx_data, 64,0)<0){
+		ErrorHandler("Error receiving message by FDCAN %d", instance->fdcan_number);
+		return false;
+	}
 
-	data->identifier = header_buffer.Identifier;
-	data->data_length = static_cast<FDCAN::DLC>(header_buffer.DataLength);
+	data->identifier = (data->rx_data[0] | data->rx_data[1] | data->rx_data[2] | data->rx_data[3]);
+	data->data_length = static_cast<FDCAN::DLC>((data->rx_data[4] | data->rx_data[5]| data->rx_data[6] | data->rx_data[7]));
+	std::copy(data->rx_data.begin() + 8, data->rx_data.end(), data->rx_data.begin());
```
g0nz4I0, line 139: I think this is undefined behaviour, 


Check cppreference: [https://en.cppreference.com/w/cpp/algorithm/copy](https://en.cppreference.com/w/cpp/algorithm/copy)
I quote:

> Copies all elements in the range [first, last) starting from first and proceeding to last.
>  If d_first is in [first, last), the behavior is undefined. In this case, [std::copy_backward](https://en.cppreference.com/w/cpp/algorithm/copy_backward) may be used instead.

I think the easiest thing is to create an auxiliary reception buffer and just copy from there, unparsing the id, the DLC and the data. Besides from this, good job!

Inc/HALALMock/Services/Communication/FDCAN/FDCAN.hpp
```diff
@@ -58,18 +59,17 @@ class FDCAN{
      *        predefined instances should be used.
      *
      */
+    uint8_t Port_counter; //Puede ser inutil, por qie fdcam_number ya esta, pero bueno por si acaso
```
jmaralo, line 62:
```suggestion
    uint8_t port_counter;
```
no need for the comment, it is already being used

Src/HALALMock/Services/Communication/FDCAN/FDCAN.cpp
```diff
@@ -43,40 +53,33 @@ void FDCAN::start(){
...
-		header.Identifier = 0x0;
-
-		instance->tx_header = header;
 
 		instance->rx_queue = queue<FDCAN::Packet>();
 		instance->tx_data = vector<uint8_t>();
 
-		if(HAL_FDCAN_Start(instance->hfdcan) != HAL_OK){
-			ErrorHandler("Error during FDCAN %d initialization.", instance->fdcan_number);
+		instance -> socket = socket(AF_NET,SOCK_STREAM,0);
```
jmaralo, line 60:
```suggestion
		instance -> socket = socket(AF_NET,SOCK_DGRAM,0);
```
otherwise it is TCP, and to do broadcast you need to use UDP

Src/HALALMock/Services/Communication/FDCAN/FDCAN.cpp
```diff
@@ -43,40 +53,33 @@ void FDCAN::start(){
...
+		if(instance -> socket < 0){
+			ErrorHandler("Error creating socket for FDCAN %d", instance->fdcan_number);
 		}
+		struct sockaddr_in BroadcastAddress;
+		BroadcastAdress.sin_family = AF_INET;
+		BroadcastAdress.sin_port = FDCAN_PORT_BASE + Port_counter;
+		BroadcastAdress.sin_addr.s_addr = INADDR_ANY;
```
jmaralo, line 67: let the user specify this IP in a config file by adding an extern variable, something like
```c++
extern const std::string fdcan_ip_address;
```
I said it was a std::string, but I don't know the exact type that sin_addr.s_addr expects

Src/HALALMock/Services/Communication/FDCAN/FDCAN.cpp
```diff
@@ -89,43 +92,52 @@ bool FDCAN::transmit(uint8_t id, uint32_t message_id, const char* data, FDCAN::D
...
+    temp_data[0] = (message_id >> 24);
+    temp_data[1] = (message_id >> 16);
+    temp_data[2] = (message_id >> 8);
+    temp_data[3] = message_id;
 
-	HAL_StatusTypeDef error = HAL_FDCAN_AddMessageToTxFifoQ(instance->hfdcan, &instance->tx_header, (uint8_t*)data);
+	temp_data[4] = ((uint32_t)dlc >> 24);
+	temp_data[5] = ((uint32_t)dlc >> 16);
+	temp_data[6] = ((uint32_t)dlc >> 8);
+	temp_data[7] = (uint32_t)dlc;
```
jmaralo, line 106: you should explicitly cast to char, and also `static_cast` over the usual C-style casts

Src/HALALMock/Services/Communication/FDCAN/FDCAN.cpp
```diff
@@ -89,43 +92,52 @@ bool FDCAN::transmit(uint8_t id, uint32_t message_id, const char* data, FDCAN::D
 		return false;
 	}
 
-	instance->tx_header.Identifier = message_id;
+    size_t buffer_len = sizeof(message_id)+ sizeof(dlc) + dlc;
```
jmaralo, line 95: dlc is an enum and the variant values problably aren't 4 for BYTES4, 8 for BYTES8, etc. Create a new function `ssize_t dlc_to_number_of_bytes(FDCAN::DLC dlc)` which returns this number and inside do a switch:
```c++
switch (dlc) {
    case DLC::BYTES4:
        return 4;
    case DLC::BYTES8:
        return 8;
    // rest of cases
```

Src/HALALMock/Services/Communication/FDCAN/FDCAN.cpp
```diff
@@ -89,43 +92,52 @@ bool FDCAN::transmit(uint8_t id, uint32_t message_id, const char* data, FDCAN::D
...
+    temp_data[3] = message_id;
 
-	HAL_StatusTypeDef error = HAL_FDCAN_AddMessageToTxFifoQ(instance->hfdcan, &instance->tx_header, (uint8_t*)data);
+	temp_data[4] = ((uint32_t)dlc >> 24);
+	temp_data[5] = ((uint32_t)dlc >> 16);
+	temp_data[6] = ((uint32_t)dlc >> 8);
+	temp_data[7] = (uint32_t)dlc;
 
-	if (error != HAL_OK) {
+	memcpy((temp_data + sizeof(message_id)+ sizeof(dlc)), data, dlc);
```
jmaralo, line 108: again, dlc is an enum, not a value you can use as the length here, this won't work as you expect because the variant values won't match

Src/HALALMock/Services/Communication/FDCAN/FDCAN.cpp
```diff
@@ -89,43 +92,52 @@ bool FDCAN::transmit(uint8_t id, uint32_t message_id, const char* data, FDCAN::D
...
+		
+	if(send(instance->socket,temp_data, strlen(data))<0){
```
jmaralo, line 110: Might seem weird, but a single call to send might not send all data, you should place this in a loop and check how many bytes have been sent, like so:
```c++
ssize_t total_bytes_sent{0};
while (total_bytes_sent < strlen(temp_data)) {
    ssize_t bytes_sent = send(instance->socket, temp_data + total_bytes_sent, strlen(temp_data) - total_bytes_sent);
    // if (bytes_sent < 0 ...
    total_bytes_sent += bytes_sent
}
```
Seems weird if you don't know how the protocols behind it work but could happen and the documentation on send isn't very explicit about this

Src/HALALMock/Services/Communication/FDCAN/FDCAN.cpp
```diff
@@ -89,43 +92,52 @@ bool FDCAN::transmit(uint8_t id, uint32_t message_id, const char* data, FDCAN::D
...
-	HAL_FDCAN_GetRxMessage(FDCAN::registered_fdcan.at(id)->hfdcan, FDCAN::registered_fdcan.at(id)->rx_location, &header_buffer, data->rx_data.data());
+	
+	if(recv(instance->socket, data->rx_data, 64,0)<0){
+		ErrorHandler("Error receiving message by FDCAN %d", instance->fdcan_number);
+		return false;
+	}
 
-	data->identifier = header_buffer.Identifier;
-	data->data_length = static_cast<FDCAN::DLC>(header_buffer.DataLength);
+	data->identifier = (data->rx_data[0] | data->rx_data[1] | data->rx_data[2] | data->rx_data[3]);
```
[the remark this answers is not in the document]
jmaralo: yup, missing some `<<`s

jmaralo (changes requested):
main thing is a weird "feature" with send that needs to be in a loop and putting some things in a configuration file. But for the rest looks great!

Src/HALALMock/Services/Communication/FDCAN/FDCAN.cpp
g0nz4I0: this is a good point

g0nz4I0 (approved):
just for the switchon the DLC, it seems like there is a map 
`static unordered_map<FDCAN::DLC, uint8_t> dlc_to_len;`

which I didn't know existed, maybe you can just do `return dlc_to_len[dlc];`
and avoid the switch

Src/HALALMock/Services/Communication/FDCAN/FDCAN.cpp
```diff
@@ -49,6 +49,64 @@ uint8_t FDCAN::inscribe(FDCAN::Peripheral& fdcan){
...
+			break;
+		case DLC::BYTES_48:
+			number_of_bytes = 48;
+			break;
+		case DLC::BYTES_64:
+			number_of_bytes = 64;
+			break;
+		default:
+			number_of_bytes = 4; 
+			break;
```
jmaralo, line 105: don't know about this default case, might be good to rise an error handler or debug this info, but for now its ok

jmaralo (approved):
You could do return directly on the switch, but if you are to change it I like more what gonzalo proposed of using a map, still LGTM

PR #440, opened 2024-11-01
Change public to private In Communication

oganigl:
I've moved functions and variables that must be only use by the class to private so there is no lack of compatibility between halal and halmock


g0nz4I0:
this PR is interesting, but two things, I'm not sure if it will work because some of this things are called from C code in LwIp, so I'm not entirely sure if they can be made private, also the CI (Continuous Integration) check has failed, take a look so we all start learning to read CI checks/results. They will be very useful throughout the year

g0nz4I0 (changes requested):
See old comment

g0nz4I0 (approved):

PR #441, opened 2024-11-03
[FW-88] Added stuff to allow developing inside a container

jorgesg82:
This Dockerfile and  config file allow to use the "Dev Containers" extension in Visual Studio Code to develop ST-LIB inside a container, in which compiler, cmake and all required tools for compiling and developing are already installed and configured.

This only works for compiling and developing ST-LIB in simulator mode. Compiling for STM32 will be investigated in the future

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,1 @@
+FROM mcr.microsoft.com/devcontainers/cpp:1-ubuntu-24.04
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,13 @@
+{
+	"name": "C++",
+	"build": {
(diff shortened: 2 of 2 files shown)


g0nz4I0:
Nice work, and for reference this are the contents right: [mcr.microsoft.com/devcontainers/cpp:1-ubuntu-24.04](mcr.microsoft.com/devcontainers/cpp:1-ubuntu-24.04)

g0nz4I0 (approved):

jorgesg82:
All documentation about this tool is here: [](https://containers.dev). I used an image from microsoft that already includes all necessary tools that we need for developing and compiling ST-LIB in simulator mode, but this is a simple Dockerfile, if we need more control/customization in the container, we can always pull a generic ubuntu image and build all we need from that, or build from this image too.

PR #443, opened 2024-11-08
[FW-108] Added cubeclt to container

jorgesg82:
Now ST-LIB has a container to develop and compile it for Simulator and for real MCU.

## Setup
To use it you must install [Dev Containers extension on VSCode](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) and [Podman](https://podman.io) (you can try to use docker, but it hasn't been tested), and set dev containers extension to use Podman (you can see how [here](https://code.visualstudio.com/remote/advancedcontainers/docker-options#_podman)). 
Then, you must configure podman machine to increase memory (by default is 2GB) to 4 - 8 GB (the more the better). This step can be done through the GUI or using command line like this: 
```sh
podman machine stop
podman machine set --cpus 2 --memory 4096
podman machine start
```


g0nz4I0 (approved):

jmaralo (approved):
I know intellisense doesnt work, but I dont see why this shouldn't work

PR #445, opened 2024-11-12
[FW-82] Better IP and ports declaration in SPI mock

jorgesg82:
I just change ip and ports declaration so now port is an attribute of Instance, and VMCU IP has a better name.


jorgesg82:
SPI can now have his own IP, which can be different from the FDCAN IP

g0nz4I0 (approved):

PR #451, opened 2024-11-14
Added setup file

StefanCostea:
I think the addition of a setup file is necessary to ensure the correct environment setting of the workspace now that things are going to start to run natively, I just created the template file


jmaralo (approved):
True, nice addition

PR #457, opened 2024-11-17
[FW-95]TestEncoder

Ikubimu:
Passed test Inscribe, turn_on, turn_off, get_direccion, get counter and reset.

Inscribe on Encoder.cpp has been modifiied to return a id != 0


Src/HALALMock/Services/Encoder/Encoder.cpp
```diff
@@ -22,7 +23,7 @@ uint8_t Encoder::inscribe(Pin &pin1, Pin &pin2) {
             pin1.to_string().c_str(), pin2.to_string().c_str());
         return 0;
     }
-    uint8_t id = Encoder::id_counter++;
+    uint8_t id = ++Encoder::id_counter;
```
g0nz4I0, line 26: why

Src/HALALMock/Services/Encoder/Encoder.cpp
```diff
@@ -22,7 +23,7 @@ uint8_t Encoder::inscribe(Pin &pin1, Pin &pin2) {
             pin1.to_string().c_str(), pin2.to_string().c_str());
         return 0;
     }
```
g0nz4I0, line 25: I think you need to do id_counter-- here

g0nz4I0 (changes requested):
I understand the logic behind returning 0 if it fails, but there is no strict need, the ErrorHandler will trigger.

You can check it was triggered by checking a  variable `static double ErrorHandler::error_triggered;`

g0nz4I0 (approved):

jdmarmen (approved):

PR #461, opened 2024-11-18
[FW-94] Changes in ADC + included tests

jdmarmen:
- Changes in ADC: 

1. The original implementation wasn't accessing the pins memory location, it was creating a copy. Changed the implementation to work with pointers.
2. Added the pin type selection inside inscribe(), as it wasn't done.
3. Added a map to pair each ADC pin ID with it's corresponding pin.
4. More changes on inscribe() start() and get_value_pointer() to support the new implementation.
5. Small changes on SharedMemory::get_pin().

- Added ADC tests


Tests/ADCTest.cpp
```diff
@@ -0,0 +1,78 @@
...
+TEST(ADC,Get_value){
+
+    SharedMemory::start("GPIO_Name", "State_Machine_Name");
+
+    uint8_t adc_id=ADC::inscribe(PA0);
+
+    ADC::start();
+    ADC::turn_on(adc_id);
+
+    uint16_t* my_raw_value=ADC::get_value_pointer(adc_id);
```
gonzalo-sm, line 51: I dont understand this line

gonzalo-sm (approved):

g0nz4I0 (approved):

jorgesg82 (approved):

PR #462, opened 2024-11-19
[FW-98] PWM Tests

jdmarmen:
- Small changes in set_dead_time to ensure that we can only set the dead_time if the PWM is off.
- Added PWM tests


gonzalo-sm (approved):
Doesn't this trigger an ErrorHandler? because I dont see a pin map in this PR defining the map with `available_pwms`

gonzalo-sm (changes requested):
I'm disapproving it, Check ErrorHandler is not triggered

gonzalo-sm:
Add the checks to ErrorHandler in the implementation of the class, both in PWM and DualPWM, then check that ErrorHandler is not triggered in the test.

g0nz4I0 (approved):

Ikubimu (approved):

PR #463, opened 2024-11-19
[FW-99] DualPWM Tests

jdmarmen:
- Modified DualPWM structure for accessing emulated pin variables, now the Emulated Pins are instance members and get initialized in the constructor to modify their variable values when calling set_duty_cycle(), set_frequency() and set_dead_time(). In the old version, we were only able to modify the positive pin value, the negative one could only be accessed in the constructor.
- Added DualPWM tests.


gonzalo-sm (changes requested):
Same as in the other PR, this test should actually fail, because an ErrorHandler will be triggered

Ikubimu (approved):
It's seems good

g0nz4I0 (changes requested):
Remove the ErrorHandlerTriggered check in all the functions, we only want the check inside the test, not in the library function

g0nz4I0 (approved):

PR #467, opened 2024-11-20
Adapt CMakeLists for better integration with projects

jmaralo:
Checks if the variable `STLIB_NAME_SUFFIX` is defined when running the file and slightly alters the names so that a project can compile multiple ST-LIBs with different compile options (ethernet vs no ethernet for example)


g0nz4I0 (commented):
Why did you remove the enable testing call?

jmaralo:
Because it is already called in `tests/CMakeLists.txt` file

jdmarmen (approved):

g0nz4I0 (approved):

PR #471, opened 2024-11-26
[FW-119] Improve SPI

jorgesg82:
SPI has been modified to use a command protocol that consists on two commands: SLAVE_SELECT and SLAVE_DESELECT. The first one is sent under the number 1, while the second one is sent under the number 2. If a normal package is sent, it must be preceded by a zero to indicate that there isn't a command and it shouldn't be processed as it.


g0nz4I0 (approved):
multithreaded PR, probably just merge and check for errors while using it

jmaralo (approved):
Small PR with some minor fixes I see, LGTM. (also what gonzalo said)

PR #474, opened 2024-12-05
[FW-137] Add turn_on() method to PWMSensor

jdmarmen:
PWMSensor inscribed an InputCapture pin but did not include the call to the turn_on() method, so it was not working.

### DIFF
--- a/Inc/ST-LIB_LOW/Sensors/PWMSensor/PWMSensor.hpp
+++ b/Inc/ST-LIB_LOW/Sensors/PWMSensor/PWMSensor.hpp
@@ -28,12 +28,14 @@ template<class Type>
 PWMSensor<Type>::PWMSensor(Pin &pin, Type &frequency, Type &duty_cycle) :
 	frequency(&frequency), duty_cycle(&duty_cycle) {
 		id = InputCapture::inscribe(pin);
+		InputCapture::turn_on(id);
 }
 
 template<class Type>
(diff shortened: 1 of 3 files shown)


g0nz4I0 (changes requested):
I dont think this works, can you add some tests to PWMSensor, if I remember correctly turn_ons must be performed **AFTER** `STLIB::start()`, and the inscribe needs to be performed before `STLIB::start()` so this solution is incompatible. 

Again, try to make a test and check it

oganigl:
I think the best idea might be two methods to turn_on and turn_off the pwm_sensor. So you have control of the pwm sensor and you can construct before but do the turn on after

jdmarmen:
> I dont think this works, can you add some tests to PWMSensor, if I remember correctly turn_ons must be performed **AFTER** `STLIB::start()`, and the inscribe needs to be performed before `STLIB::start()` so this solution is incompatible.
> 
> Again, try to make a test and check it

I've done a small the test with the python and works correctly, the sensor is declared before start()

g0nz4I0 (changes requested):
I'm still not approving it, I looked at it further and it should be done like this;

image

See that LinearSensor does not have a turn_on method, yet the ADC does and it gets called, so its just a matter of calling it at the same time, this way we are consistent,

jmaralo (changes requested):
Creating a sensor shouldn't turn it on, it should be a separate action. More or less what gonzalo already said

g0nz4I0 (approved):

PR #479, opened 2024-12-26
Fw-136-Code-generation-for-data-packets

Cantonplas:
Python script that creates the .hpp for the data packets based on the .json files from the JSON.ADE directory. It also has description classes to make it more easy for more code generation based on this script. Works quite good,  it has an example attached that has been autogenerated.


jmaralo:
This should be in the template project, not the ST-LIB

Inc/Packet_generation/DataPackets.hpp
```diff
@@ -0,0 +1,39 @@
...
...6_t idpacket315,general_state general_state,slave_state slave_state,float32 lcu_coil_temp_1,float32 lcu_coil_temp_2,float32 lcu_coil_temp_3,float32 lcu_coil_temp_4,float32 lcu_coil_temp_5,float32 lcu_coil_temp_6,float32 lcu_coil_temp_7,float32 lcu_coil_temp_8,float32 lcu_coil_temp_9,float32 lcu_coil_temp_10,float32 lcu_driver_temp_1,float32 lcu_driver_temp_2,float32 lcu_driver_temp_3,float32 lcu_driver_temp_4,float32 lcu_driver_temp_5,float32 lcu_driver_temp_6,float32 lcu_driver_temp_7,float32 lcu_driver_temp_8,float32 lcu_driver_temp_9,float32 lcu_driver_temp_10,float32 lcu_coil_current_1,float32 lcu_coil_current_2,float32 lcu_coil_current_3,float32 lcu_coil_current_4,float32 lcu_coil_current_5,float32 lcu_coil_current_6,float32 lcu_coil_current_7,float32 lcu_coil_current_8,float32 lcu_coil_current_9,float32 lcu_coil_current_10,float32 lcu_vbat_1,float32 lcu_vbat_2,float32 lcu_vbat_3,float32 lcu_vbat_4,float32 lcu_vbat_5,float32 lcu_vbat_6,float32 lcu_vbat_7,float32 lcu_vbat_8,float32 lcu_vbat_9,float32 lcu_vbat_10,float32 lcu_airgap_1,float32 lcu_airgap_2,float32 lcu_airgap_3,float32 lcu_airgap_4,float32 lcu_airgap_5,float32 lcu_airgap_6,float32 lcu_airgap_7,float32 lcu_airgap_8)
```
jmaralo, line 28: the constructor should take references, like `uint16_t &idpacket316`, `uint32_t &current_control_frequency`, and so on, otherwise it wont work as expected

Inc/Packet_generation/Generator.py
```diff
@@ -0,0 +1,191 @@
...
+        enum = enum.replace("%values%", values.strip())
+        return enum
+    
+    @staticmethod
+    def _unsigned_int_correction(type:str):
+        aux_type = type[:4]
+        if aux_type == "uint":
+            type += "_t"
+        return type
+
```
jmaralo, line 91: since we are going to reuse this logic in multiple places, it would be great if you could split this file into two, one which parses the documents and another one that fills the template

jmaralo (changes requested):

PR #489, opened 2025-02-01
FW[240] Added shm count to the state machine default constructor

jdmarmen:


Inc/ST-LIB_LOW/StateMachine/StateMachine.hpp
```diff
@@ -120,7 +120,7 @@ class StateMachine {
 	bool is_on = true;
 
 
-	StateMachine() = default;
+	StateMachine();
```
jorgesg82, line 123: You have erased `default` , but you are only implementing the constructor if SIMON is defined. This will work if you are not building for simulator?

jorgesg82 (changes requested):

Inc/ST-LIB_LOW/StateMachine/StateMachine.hpp
jdmarmen: What you are saying makes sense, nevertheless, all the actions passed correctly, and that includes building every option.  Just to make sure, i added an else to the ifdef in case we aren't compiling with SIMON.

jmaralo: should be the same but there are some [minor differences](https://www.reddit.com/r/cpp_questions/comments/14wu8gn/comment/jrjosoz/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button)

jmaralo (approved):

oganigl (approved):
lgtm

PR #490, opened 2025-02-03
Fixes on FDCAN for BCU

jmaralo:
This PR fixes some bugs with FDCAN found when working with the BCU. Main issue was that the FDCAN interrupt callback was consuming the packets constantly, so no packets made it to the user code or only made it on rare occasions.


jmaralo:
Sorry for the linter

StefanCostea:
This PR downgrades CAN FD to CAN, if needed CAN aside, do not change FDCAN module rather create a new one, FDCAN and CAN are not compatible

StefanCostea:
Most probably packets were being consumed because of the INTERNAL LOOPBACK configuration which does not allow packets to exit the MCU, commenting the callback will not fix this, the other way around, it breaks existing packet class for CAN

jmaralo:
> This PR downgrades CAN FD to CAN, if needed CAN aside, do not change FDCAN module rather create a new one, FDCAN and CAN are not compatible

True, I'll make another class for it

jmaralo:
> Most probably packets were being consumed because of the INTERNAL LOOPBACK configuration which does not allow packets to exit the MCU, commenting the callback will not fix this, the other way around, it breaks existing packet class for CAN

The callback was clearly causing issues because there was no way for the application code to read the packet.

At least what I got from the FDCAN class public facing methods the idea is to call read method in a loop, if the read method is already called in the interrupt callback, the latest packet gets consumed, and the function always returns false because the read method checks if there is new data (there is no new data because it has already read the latest data)

StefanCostea:
> > Most probably packets were being consumed because of the INTERNAL LOOPBACK configuration which does not allow packets to exit the MCU, commenting the callback will not fix this, the other way around, it breaks existing packet class for CAN
> 
> The callback was clearly causing issues because there was no way for the application code to read the packet.
> 
> At least what I got from the FDCAN class public facing methods the idea is to call read method in a loop, if the read method is already called in the interrupt callback, the latest packet gets consumed, and the function always returns false because the read method checks if there is new data (there is no new data because it has already read the latest data)

You are right in this, I did not know what the read method was doing under the hood, read actually is a very poor name, retrieve will be more accurate

g0nz4I0:
So we wait and make another class for FDCAN, or we can at runtime set the configuration, without having to add a diferent class, based on some binary flags to set the use of extended id, bit rate switching and extended data frame.

g0nz4I0:
I attach images showing that the various bitrate configurations work


The new template implementation is used something similar to this:
```c++
auto can_id = FDCAN::inscribe<CANBitRatesSpeed::CAN_1_Mbit,CANFormat::CAN_NORMAL_FORMAT,
                              CANIdentifier::CAN_11_BIT_IDENTIFIER,CANMode::CAN_MODE_NORMAL>(FDCAN::fdcan1);
```

The test code is the following, compiled under Board[DBG] in Windows system, note NO Ethernet


```c++
int main(void) {
    auto can_id = FDCAN::inscribe<CANBitRatesSpeed::CAN_1_Mbit,CANFormat::CAN_NORMAL_FORMAT,
            CANIdentifier::CAN_11_BIT_IDENTIFIER,CANMode::CAN_MODE_NORMAL>(FDCAN::fdcan1);
    STLIB::start();
     char data[8] = {'1','_','_','_','M','B','i','t'};
    Time::register_low_precision_alarm(5, [&]() { 
        FDCAN::transmit(can_id,0x55,&data[0],FDCAN::BYTES_8);
    });

    while (1) {
        STLIB::update();
    }
}
```
1MBit
125kBit
250kBit

oganigl (commented):
For me looks good the template FDCAN,  I trust the assigns to fdcan_instance are good

oganigl (approved):

PR #494, opened 2025-02-16
[FW-248] Added capture in Heap and Stack orders

jorgesg82:
Heap and Stack orders have been modified to accept a std::function instead of a pointer to a function, allowing to capture in lambda expressions.


g0nz4I0 (commented):
On paper it looks good, have you tested it with real HW? I'm not sure if it can be invoked from C code. Also I believe the check may pass because it is a template, so it is not getting instantiated.

jorgesg82:
I think I checked it on HVSCU, but I am not sure, I can assure it tomorrow

PR #495, opened 2025-02-16
[Fw-248] Heap and Stack orders capture

jorgesg82:
Heap and stack orders doesn't allow to capture anything in lambda expressions, as it accept a pointer to a function, instead of an std::function. This PR change that

### DIFF
--- a/Inc/HALAL/Models/Packets/Order.hpp
+++ b/Inc/HALAL/Models/Packets/Order.hpp
@@ -11,7 +11,7 @@
 
 class Order : public Packet{
 public:
-    virtual void set_callback(void(*callback)(void)) = 0;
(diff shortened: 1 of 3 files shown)


g0nz4I0 (commented):
On paper it looks good, but have you tested it with real HW? I'm not sure if you can call it from C code, I believe it may pass the checks because it is a template and it is not getting instantiated. Be sure to double check that

StefanCostea:
Just be aware this was a purposful design desition, to always have a static function that is beign referenced at least I would make sure just r value reference functions can be set as a callback, to ensure they will always have a value when settign the callback, additionally using STL functional heavily impacted the performance, I would recommend using an alternative using a template type with some is_callable concept requirement.

PR #496, opened 2025-02-21
Change direction from double to bool

oganigl:


jmaralo (approved):

Ikubimu (approved):
wow 8bytes for a simple bool

PR #497, opened 2025-02-25
[FW-339] Fix SPI

jorgesg82:
This bug has been detected when trying to use the old BMS-LIB. It was working with the old ST-LIB but it ins't working with the new one. Comparing both, It has been detected that two functions were removed: receive and transmit_and_receive without DMA. It has been tried to add them again and BMS-LIB is working.

@Ikubimu wasn't able neither to use SPI, apart of SPIOrder (which use HAL directly), so maybe with this change now he is able to use it.

Is important that @g0nz4I0 or @jmaralo review this, as I guess this code was removed for some reason that the new ones can't figure out, so please, give us your experienced opinion


StefanCostea:
Most probably the non DMA variant was removed due to its blocking nature. This addition is fine, however I would recomend refactoring the BMS-LIB to use the non blocking variant

jmaralo (approved):

jorgesg82:
> Most probably the non DMA variant was removed due to its blocking nature. This addition is fine, however I would recomend refactoring the BMS-LIB to use the non blocking variant

Yes, the idea is to remake the BMS-LIB this year. I want even to try to make it STLIB-independent. But, at the moment, we should use the old one, and it doesn't use DMA, so I need this. Actually, It isn't necessary to merge that on master, I just need this branch so I can point to this from BMS-LIB

Ikubimu (approved):

PR #499, opened 2025-03-19
Fix boundary to respect logical priorities

jorgesg82:
A check in protections wasn't respecting logical priorities, as C++ doesn't respect them by default, because `and` operation has the most priority. This results as a warning when compiling, preventing compilation with warnings as errors.


g0nz4I0 (approved):
I don't know how it would be misinterpreted by the compiler though, but it fixes it so go ahead

jmaralo (approved):

PR #500, opened 2025-03-23
Different encoder sensor algorithm

jmaralo:
I tried to move around a couple things in hopes of improving how the encoder works. Apart from that the PR includes a small fix on a reorder warning I got on the cabinet HVSCU.

I haven't tested it yet, but I will try to do so tomorrow.

**TL;DR: Call the read method of the encoder sensor in a cyclic action!**

Here is a long rant on how the encoder currently works.

The main feature on top of the regular Encoder is that the EncoderSensor calculates the position, speed and acceleration from the Encoder counter. To do so it has to perform two derivatives.

The position is simple, subtract the current counter value from the starting one and multiply by the meters each counter step represent.

Now, for the speed and acceleration, there is a neat trick to calculate both with three position samples equally spaced apart (for example the position now, the position 10ms ago and the position 20ms ago). I don't properly understand the math, but @ricardochust told me about it and it made sense. It basically does some operations on the three positions and has less error than taking just two points and making two derivatives on it.

Each time the read method is called this process is repeated and it EXPECTS THAT THE METHOD IS CALLED IN REGULAR TIME INTERVALS (inside a cyclic action). The previous version had a protection but I decided to remove it for now.

Internally the EncoderSensor stores multiple position samples in a ring buffer and uses the last and mid-point of the ring buffer to do the calculations, this is because if the points are too close by it could lead to errors, but maybe the refresh rate has to be high, so maybe you want to take points spaced 500ms away but refresh the position, speed and acceleration each 10ms, to do so you would need to call the read method each 10ms and set the EncoderSensor samples to 100 (half point is 50 * 10ms -> 500ms in the past).

If this didn't make any sense I understand :)


g0nz4I0 (approved):
LGTM

g0nz4I0 (changes requested):
I approved earlier but I have realized that it drops the convention across the library of using output parameters when using sensors, add that interface and it is good to go. We love/hate legacy code.

g0nz4I0 (approved):

Src/HALAL/Services/Encoder/Encoder.cpp
```diff
@@ -146,12 +146,16 @@ void Encoder::init(TimerPeripheral* encoder) {
...
+                                 uint64_t last_clock_time) {
+    int64_t delta_clock = clock_time - last_clock_time;
+    if (clock_time < last_clock_time) {  // overflow handle
+        delta_clock =
+            clock_time +
+            CLOCK_MAX_VALUE * NANO_SECOND / HAL_RCC_GetPCLK1Freq() * 2 -
+            last_clock_time;
+    }
+    return delta_clock;
+}
```
g0nz4I0, line 159: Linter changes seems

Src/HALAL/Services/Encoder/Encoder.cpp
```diff
@@ -86,10 +86,10 @@ void Encoder::reset(uint8_t id) {
...
 
 uint32_t Encoder::get_counter(uint8_t id) {
-    // if (not Encoder::registered_encoder.contains(id)) {
-    // 	ErrorHandler("No encoder registered with id %u", id);
-    // 	return 0;
-    // }
+    if (not Encoder::registered_encoder.contains(id)) {
+        ErrorHandler("No encoder registered with id %u", id);
+        return 0;
+    }
```
g0nz4I0, line 92: remove this altogether, it is not worth to perform this check in the hot-loop every time

g0nz4I0 (changes requested):
Address mentioned comments

oganigl (dismissed):
Looks good to me

Src/HALAL/Services/Encoder/Encoder.cpp
jmaralo: yes

Inc/ST-LIB_LOW/Sensors/EncoderSensor/EncoderSensor.hpp
```diff
@@ -7,40 +7,80 @@
...
+
+    const double counter_distance_m;
+    const double sample_time_s;
+
+    uint8_t encoder_id;
+
+    // We want to get the last buffer element and the midpoint, if the number of
+    // elements is odd, these points and the present won't be evenly spaced
+    // across time. The SAMPLES computation rounds it down to make it even
+    RingBuffer<int64_t, (SAMPLES / 2) * 2> past_delta_counters{};
```
g0nz4I0, line 29: Fun fact: you can make this check with a requires in the template declaration

g0nz4I0 (approved):

Inc/ST-LIB_LOW/Sensors/EncoderSensor/EncoderSensor.hpp
jmaralo: probably, but this way also works and is almost transparent to the user, they don't need to know how the ring buffer works

PR #502, opened 2025-03-28
Add pin lock method to digital out

jmaralo:
The HAL has a method that can lock a GPIO in a certain state until the board resets (using HAL_GPIO_LockPin). This could be useful for the contactors as an example, where once you enter fault you don't want to close them until you reset the board after fixing the fault cause. This is the same for the GD reset and enable, brakes, etc. All driven by a GPIO.

Also I took the chance to run the linter through the digital out code, the logic of this modification is minimal. The interesting methods are named force_lock_state


Ikubimu (approved):
lgfm

g0nz4I0 (approved):
I'm going to shove the linter up *********

PR #503, opened 2025-04-16
Center Aligned

oganigl:
To center Aligned the pwms of a timer if we use phased instead of Normal. Added a boolean to check if is phased and multiply the freq to don't change the value


g0nz4I0 (commented):
Why are you using the PHASED setting, shouldn't you be creating a new setting, PHASED for this purpose?

oganigl:
@g0nz4I0  Yes I used the oscilloscope to tested, If I understand your comment, your complaining about the boolean to change a single PWM to phased??

Src/HALAL/Models/TimerPeripheral/TimerPeripheral.cpp
```diff
@@ -111,8 +113,7 @@ void TimerPeripheral::init() {
 			sConfigOC.OCNIdleState = TIM_OCNIDLESTATE_RESET;
 
 			if (pwm_data.mode == PHASED) {
-				//ASSYMETRIC_MODE_1 means one output per pair of registers (CCR1 - CCR2) for example
-				sConfigOC.OCMode = TIM_OCMODE_ASSYMETRIC_PWM1;
+				sConfigOC.OCMode = TIM_OCMODE_PWM1;
```
jmaralo, line 116: why are you changing this?

Src/HALAL/Models/TimerPeripheral/TimerPeripheral.cpp
```diff
@@ -47,17 +47,19 @@ void TimerPeripheral::init() {
...
+        handle->Init.ClockDivision = TIM_CLOCKDIVISION_DIV1;
+        handle->Init.RepetitionCounter = 0;
+        handle->Init.AutoReloadPreload = TIM_AUTORELOAD_PRELOAD_DISABLE;
+        for (PWMData pwm_data : init_data.pwm_channels) {
 			if (pwm_data.mode == PHASED) {
-				handle->Init.CounterMode = TIM_COUNTERMODE_CENTERALIGNED1;
-				break;
+				handle->Init.CounterMode = TIM_COUNTERMODE_CENTERALIGNED3;
+				handle->Init.AutoReloadPreload = TIM_AUTORELOAD_PRELOAD_ENABLE;
+			    break;
```
jmaralo, line 58: why are you changing this, you have a new CENTER_ALINGED mode, if this is for center aligned change it there

jmaralo (changes requested):
there are a few changes you made that make no sense, if they are to support center aligned PWMs you should add a new if instead of modifying the PHASED configuration

jmaralo:
Also add support for center aligned dual PWMs

Src/HALAL/Models/TimerPeripheral/TimerPeripheral.cpp
oganigl: yes, sorry I forgot to changed in the timerPeripheral when I added the new mode

Src/HALAL/Models/TimerPeripheral/TimerPeripheral.cpp
oganigl: same than before

Src/HALAL/Models/TimerPeripheral/TimerPeripheral.cpp
```diff
@@ -47,17 +47,21 @@ void TimerPeripheral::init() {
 		handle->Instance = handle_to_timer[handle];
 		handle->Init.Prescaler = init_data.prescaler;
 		handle->Init.CounterMode = TIM_COUNTERMODE_UP;
-		for (PWMData pwm_data : init_data.pwm_channels) {
+		handle->Init.AutoReloadPreload = TIM_AUTORELOAD_PRELOAD_DISABLE;
+        for (PWMData pwm_data : init_data.pwm_channels) {
```
g0nz4I0, line 51: Dont make a copy, use a reference

g0nz4I0 (approved):

jmaralo (approved):

g0nz4I0 (approved):

PR #504, opened 2025-04-16
Class_DWT

oganigl:
Added class to count the number of clocks of a function (pijada maxima)


g0nz4I0:
Please provide a description on how this class would be used and for what purpopse

g0nz4I0:
How should I use it? How do I get the seconds elapsed or things like that? 
Honestly I would like this to be some steps towards a small benchmarking toolkit

g0nz4I0 (approved):

PR #506, opened 2025-05-03
Added method to LinearSensor to set slope

jorgesg82:
A method has been added to LinearSensor class to be able to modify the slope/gain. 

The reason to include it has been to be able to characterize sensors in runtime.


oganigl (approved):
Looks good for me

PR #512, opened 2025-05-14
Protections enhancements

g0nz4I0:
Sorry for the incredible amount of duplicated code, runtime polymorphism and some other abnormalities.

I basically added a frozen value so that that value is sent when the message is issued, this way we ensure integrity with the value that triggered the protection, this is achieved my storing a copy of the value pointed by the pointer.

I also fixed some discrepancies between the high frequency and low frequency protections checks.

I also added individual time tracking for the notification of ProtectionMessages, as otherwise only one would be sent.

I removed some code with test_fault that in the past seemed to indicate whether the Fault was coming from TCP or the board itself, because the naming wasn't clear and the Backend was doing weird stuff with Transport module and connections being closed I completely removed that part. This feature might be interesing to bring back in the future, but with more time and though put into it.

Added new functionality regarding values coming back to normal values, even while the board stays in a Fault state if the variable gets back to a normal state that should be shown in the GUI, as it is no longer dangerous.

This idea was only implemented when Warnings were used, now it is also working with only Fault and OK. I attach some screenshots showing the messages on the GUI.

image
I haven't tested the Warning parts. But it works for FAULTs and OKs
```c++
#define TESTING_WARNINGS 0
int current_2 = 0;

int main(void) {
#ifdef SIM_ON
    SharedMemory::start();
#endif
        int current = 0;
    #if TESTING_WARNINGS
        add_protection(&current,Boundary<int,ABOVE>(5,10));
        add_protection(&current,Boundary<int,BELOW>(-5,-10));
        add_protection(&current,Boundary<int,OUT_OF_RANGE>(-20,-15,15,20));
    #else
        add_protection(&current,Boundary<int,OUT_OF_RANGE>(-20,20));
        add_protection(&current_2,Boundary<int,ABOVE>(10));
        add_protection(&current,Boundary<int,BELOW>(-10));
    #endif
    ProtectionManager::initialize();

    STLIB::start();
    ServerSocket control_station{"192.168.1.4",50500};
    Time::register_low_precision_alarm(1000,[&current](){
        current++;
        current_2++;
        if(current > 30 || current_2 > 30)
        {
            current = -25;
            current_2 = 5;
        }
    });
    StateMachine sm{1};
    sm.add_state(0);
    ProtectionManager::link_state_machine(sm,0);


    while (1) {
        STLIB::update();
        ProtectionManager::check_protections();
    }
}
```

For reference I attach the code that I have used to test, 
template_project branch: main
STLIB: branched from latest development

### DIFF
--- a/Inc/ST-LIB_HIGH/Protections/Boundary.hpp
+++ b/Inc/ST-LIB_HIGH/Protections/Boundary.hpp
@@ -107,10 +107,10 @@ struct Boundary<Type, BELOW> : public BoundaryInterface{
 				warn_message = new HeapOrder(uint16_t{2000},&format_id,&boundary_type_id,&name,&this->warning_threshold,this->src,
 				&Global_RTC::global_RTC.counter,&Global_RTC::global_RTC.second,&Global_RTC::global_RTC.minute,
 				&Global_RTC::global_RTC.hour,&Global_RTC::global_RTC.day,&Global_RTC::global_RTC.month,&Global_RTC::global_RTC.year);
-				ok_message = new HeapOrder(uint16_t{3000},&format_id,&boundary_type_id,&name,&this->warning_threshold,this->src,
+			}
+			ok_message = new HeapOrder(uint16_t{3000},&format_id,&boundary_type_id,&name,&this->warning_threshold,this->src,
 				&Global_RTC::global_RTC.counter,&Global_RTC::global_RTC.second,&Global_RTC::global_RTC.minute,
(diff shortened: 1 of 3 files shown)


jorgesg82 (dismissed):
LGTM, taking into account that is hard to detect an issue only reading this kind of code hehe. But it seems fine

jorgesg82:
Ah, and I love the new feature that notifies you when the value gets in a safe range again

jmaralo (approved):

jorgesg82 (approved):
LGTM

PR #526, opened 2025-09-26
Benchmarking

g0nz4I0:
* Added SEGGER_RTT files to project
* Added SEGGER_RTT files to CMakeLists.txt
* Fixes to DataWatchpointTrace module
* IFDEFs guards for Benchmarking mode

Benchmarking Features

Different types of Events:
* For measuring time executions of code blocks
* Measuring time between different executions

This is achieved by a set of macros, which record the current CycleCounter at a given point.

Under the hood, the macros use a string which is used by the frontend to display pretty messages.
To communicate with the host PC it uses SEGGER_RTT protocol, so it needs to have a PC hooked at all times.
It introduces breaking changes by changing the software used to flash, moving away from STLINK to OPENOCD.

It also adds some polling checks to detect overflow and update the global counter.


victor-Lopez25:
Would it be possible to add in the DWT CPI counter to get info on additional cycles spent on IF stalls or multicycle instructions?
It would be better to get info on the amount of cache misses in a block of code too but it doesn't look like there's any way in STM32H723.
The CPI counter is only 8 bits but you can enable an overflow event.

g0nz4I0:
The DWT peripheral can do a bunch of stuff yes, but I havent been able to make it work with VSCode yet, the event is not a SW event, but rather an internal event that triggers a transfer via SWD, this works fine in STM32CubeIDE, but I havent been able to make it work in VScode or outside the IDE yet, it's a work in progress

I can say however that the cache thing is impossible, some things that are acctually possible are checking cycles spent in interrupts or sleeeping for example

Inc/HALAL/Benchmarking_toolkit/Benchmarking.hpp
```diff
@@ -0,0 +1,119 @@
...
+            __asm__ volatile("BKPT \n");
+        } else if constexpr(benchmark_type == BenchmarkType::RUNTIME){
+            uint32_t cycles = DWT->CYCCNT;
+            if constexpr( logging_type & _RTT){
+                timestamp.id = Id;
+                timestamp.cycles = cycles;
+                SEGGER_RTT_Write(0,&timestamp,sizeof(TimeStamp));
+            }
+        }
+    }
```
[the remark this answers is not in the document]
g0nz4I0: The overflow logic is implemented on the host side, to make the target measuremetns as lightweight as possible

Inc/HALAL/Benchmarking_toolkit/Benchmarking.hpp
```diff
@@ -1,124 +1,119 @@
...
+#ifdef RUNTIME_BENCHMARK
+    BenchmarkType::RUNTIME;
+    static const char BENCHMARK_TYPE_RUNTIME[] __attribute__((used)) = "BENCHMARK_RUNTIME";
+#elif defined(GDB_BENCHMARK)
+    BenchmarkType::GDB;
+    static const char BENCHMARK_TYPE_GDB[] __attribute__((used)) = "BENCHMARK_GDB";
+#else 
+    BenchmarkType::NONE;
+    static const char BENCHMARK_TYPE_NONE[] __attribute__((used)) = "BENCHMARK_NONE";
+#endif
```
victor-Lopez25: please define benchmark_type in each scope of the #if clause:
```cpp
#ifdef RUNTIME_BENCHMARK
    constexpr BenchmarkType benchmark_type = BenchmarkType::RUNTIME;
    static const char BENCHMARK_TYPE_RUNTIME[] __attribute__((used)) = "BENCHMARK_RUNTIME";
#elif defined(GDB_BENCHMARK)
    constexpr BenchmarkType benchmark_type = BenchmarkType::GDB;
    static const char BENCHMARK_TYPE_GDB[] __attribute__((used)) = "BENCHMARK_GDB";
#else 
    constexpr BenchmarkType benchmark_type = BenchmarkType::NONE;
    static const char BENCHMARK_TYPE_NONE[] __attribute__((used)) = "BENCHMARK_NONE";
#endif 
```

Inc/HALAL/Benchmarking_toolkit/Benchmarking.hpp
```diff
@@ -1,124 +1,119 @@
...
+#if defined(SD)
+    _SD |
+#else
+    _NONE |
+#endif
+#if defined(RUNTIME_BENCHMARK)
+    _RTT;
+#else
+    _NONE;
+#endif
```
victor-Lopez25: I know it looks cool but it doesn't look very clear, it could be better to write it like this:
```cpp
#if defined(RUNTIME_BENCHMARK)
#define RUNTIME_BENCHMARK_TAG  _RTT
#else
#define RUNTIME_BENCHMARK_TAG  _NONE
#endif

#if defined(SD)
inline constexpr int logging_type = _SD | RUNTIME_BENCHMARK_TAG:
#else
inline constexpr int logging_type = _NONE | RUNTIME_BENCHMARK_TAG;
#endif
```

or alternatively:
```cpp
#if defined(RUNTIME_BENCHMARK)
#define RUNTIME_BENCHMARK_TAG  _RTT
#else
#define RUNTIME_BENCHMARK_TAG  _NONE
#endif

#if defined(SD)
#define BENCHMARK_LOGGING_TYPE  _SD | RUNTIME_BENCHMARK_TAG
#else
#define BENCHMARK_LOGGING_TYPE  _NONE | RUNTIME_BENCHMARK_TAG
#endif
inline constexpr int logging_type = BENCHMARK_LOGGING_TYPE:
```

victor-Lopez25: or...:
```cpp
#if defined(RUNTIME_BENCHMARK)
#define RUNTIME_BENCHMARK_TAG  _RTT
#else
#define RUNTIME_BENCHMARK_TAG  _NONE
#endif

#if defined(SD)
#define SD_BENCHMARK_TAG  _SD
#else
#define SD_BENCHMARK_TAG  _NONE
#endif
inline constexpr int logging_type =  RUNTIME_BENCHMARK_TAG | SD_BENCHMARK_TAG:
```

PR #529, opened 2025-11-17
Feat/mdma

Cantonplas:
New class to manage the MDMA peripheral. It works on linked list mode. You can create linked list from packets, as well as merging packets to a big packet.Then you can use transfer_packet, giving the packet id that creating the packet gave you, to transfer it to a selected destination buffer or the default destination buffer you can pass as an optional parameter in the inscribe. You can also transfer data from a source buffer to a destination buffer with transfer_data.


Inc/C++Utilities/Arena.hpp
```diff
@@ -0,0 +1,179 @@
...
+template<size_t S, typename T>
+class Arena {
+   public:
+
+    /**
+     * @brief Acquire an element from the arena.
+     * @return Pointer to the acquired element, or nullptr if the arena is full.
+     */
+    T* acquire() {
+        if (freeIndexes.empty()) {
```
StefanCostea: Add unlikely

Inc/C++Utilities/Arena.hpp
```diff
@@ -0,0 +1,179 @@
...
+        return ele;
+    }
+
+    /**
+     * @brief Release an element back to the arena.
+     * @param ele Pointer to the element to release.
+     * @return True if the element was successfully released, false otherwise.
+     */
+    bool release(T* ele) {
+        if (ele < &elements[0] || ele >= &elements[S]) {
```
StefanCostea: Unlikely

StefanCostea (changes requested):
Will look more in depth tomorrow

Inc/C++Utilities/Arena.hpp
```diff
@@ -0,0 +1,179 @@
...
+    }
+    Arena(const Arena&) = delete;
+    Arena& operator=(const Arena&) = delete;
+    Arena(Arena&& other) noexcept = delete;
+    Arena& operator=(Arena&& other) = delete;
+
+   private:
+    T elements[S];
+    Stack<S> freeIndexes;
+    bool usedIndexesSet[S]{false};
```
StefanCostea: Could use a bitmap instead of a bool array then getting the first used one is as easy as ctz(used_idx_bm), loop by clearing LSB and ctz again. Just make sure you static assert that S is below the number of bits of the bitmap

Inc/HALAL/Models/MDMA/MDMA.hpp
```diff
@@ -0,0 +1,300 @@
...
+        MDMA_LinkNodeTypeDef transfer_node{};
+        uint32_t last_destination{0};
+        uint32_t last_size{0};
+        Instance() = default;
+        Instance(MDMA_HandleTypeDef handle, uint8_t id, uint8_t* data_buffer, uint8_t* destination_address,MDMA_LinkNodeTypeDef transfer_node): handle(handle), id(id), data_buffer(data_buffer), destination_address(destination_address), transfer_node(transfer_node) {}
+
+
+    };
+    static void prepare_transfer(Instance& instance, MDMA_LinkNodeTypeDef* first_node);
+    inline static std::unordered_map<uint8_t,std::vector<MDMA_LinkNodeTypeDef>> linked_lists{};
```
StefanCostea: Please avoid using map, array<uint8_t, max_expected_size> works the same

StefanCostea (changes requested):

Inc/HALAL/Models/MDMA/MDMA.hpp
```diff
@@ -0,0 +1,317 @@
...
+
+template<typename... pointers>
+inline uint8_t MDMA::add_packet(const uint8_t MDMA_id, const std::tuple<pointers...>& values)
+{
+    Instance& instance = get_instance(MDMA_id);
+    uint32_t offset{0};
+    HAL_StatusTypeDef status;
+    const uint8_t current_packet_id = number_of_packets++;
+
+    std::vector<MDMA_LinkNodeTypeDef>& nodes = linked_lists[current_packet_id];
```
StefanCostea: Again, avoid using STL container that make use of heap, use array with max size, as you receive a tuple of pointers you could even get the size from there

PR #535, opened 2025-12-08
fix: Use an XList for the domains in ST-LIB

victor-Lopez25:
Unfortunately the dummy is necessary but now we only need to specify a domain in a single place.

Had to add ## to __VA_ARGS__ for it to remove the comma when there are no varargs. This does not work in msvc unless -Zc:preprocessor is specified because their implementation of macros works differently


jorgesg82:
At the end, this feature won't work, as the MPUManager needs more flexibility. Check https://github.com/HyperloopUPV-H8/ST-LIB/pull/537

PR #536, opened 2025-12-11
Add Hardfault check

oganigl:
Add a new module to check if there was a hardfault in the last run

### DIFF
--- a/STM32H723ZGTX_FLASH.ld
+++ b/STM32H723ZGTX_FLASH.ld
@@ -35,7 +35,7 @@
 ENTRY(Reset_Handler)
 
 /* Highest address of the user mode stack */
-_sstack = ORIGIN(DTCMRAM)
+_sstack = ORIGIN(DTCMRAM);
 _estack = ORIGIN(DTCMRAM) + LENGTH(DTCMRAM);    /* end of RAM */
 /* Generate a link error if heap and stack don't fit into RAM/DTCM */
 _Min_Heap_Size = 0x200;      /* required amount of heap  */


jorgesg82:
It would also be nice to toggle the three user LEDs the Nucleo has if we are compiling targeting it

Src/HALAL/Benchmarking_toolkit/HardfaultTrace.c
```diff
@@ -0,0 +1,58 @@
...
+      for(int i = 0; i < hard_fault_leds_count;i++){
+        EnableGPIOClock(ports_hard_fault[i]);
+        InitGPIO_Output(ports_hard_fault[i],pins_hard_fault[i]);
+      } 
+}
+void Hard_fault_check(void){
+    if(*(volatile uint32_t*)HF_FLASH_ADDR == HF_FLAG_VALUE){
+        HardFaultLog log;
+        memcpy(&log,(void*)HF_FLASH_ADDR,sizeof(HardFaultLog));
+        LED_init();
```
g0nz4I0: You are not doing anything with the log info you retrieve from the flash

g0nz4I0 (changes requested):
I'll combine both reviews in one (here).

In general I like the idea, blinking leds, storage of log report, debugger detection, BUT,

There are several things I have doubts with:

1-From what I remember you can only write to flash by sectors, and also erasing whole sectors, I dont see a logic for that,


 e.g. let's say you have some code in the flash sector of you LOG report, and you HardFault, I dont think that would work as expected.


2- I really like the idea of the custom handler and the original one in assembly, but I think there might be a better idea to implement the HardFault Handler directly in assembly, otherwise maybe the stack management corrupts the PC and registers

Also consider moving the asm implementation to the library, and only exposing the C handler so that the user can define its own, in this case it doesnt really matter because we maintain both but it's good practice.


Basically my concerns are:

Flash Writing, double check that that works, maybe consider using the SD?
You are not doing anything with the log you recover from the flash (if that worked)
 Maybe you can do something with the debugger thingy and send it via UART or RTT to print the crash report 


Instead of having to call HardFaultCheck on the template project maybe you can try to take advantage of static initializtion and try to read the RST reason from RCC?
Otherwise maybe consider placing in STLIB::board, I'm curious to hear why you chose this approach

Suggestion, move the asm handler into stlib and call into a CustomHandler from the user, but this is just a comment

IMO the most important things are in the sContextFrame, which are the PC and the LR, with this you can determine the two previous calls

PR #537, opened 2025-12-13
Refactor/mpu manager

FoniksFox:
Refactor of the MPU so that it works with the new compile-infrastructure.

You can inscribe any type, on any memory domain, with any alignment (up to 32B, since there's no meaning in going any further) cacheable or non-cacheable. The MPU module will calculate and allocate on compile-time all the memory required with the required offsets and everything, such that it even fails if you request too much memory (well, the linker script is the one that actually fails).

It still gives support to the legacy MPUManager.

It has a greater range of protection of the memory, so that you cannot access any memory that is not explicitly defined as accessible.

It dinamycally (in linking-time) allocates the necessary memory with the minimum padding possible so as to not waste any unnecessary memory.

The linker script has been changed to adapt to this new structure.

You need to delete the MPUManager::start() call from Runes to make this work (since this function no longer exists).

Other changes made in the Linker Script include:
- Use DTCM (Data Tightly Coupled Memory) for Stack, Heap and Static
- Allow the use of ITCM (Instruction Tightly Coupled Memory) by declaring a function as RAM_CODE (better performance)
- Allow access to any memory region by either using an attribute or by requesting it to the MPU driver
- Optimize MPU settings for better performance
- Use Linker Script as only source of truth for the memory region start addresses and sizes
- Move some symbols that had questionable placement to the new system


Src/HALAL/Models/MPUManager/MPUManager.cpp
```diff
@@ -1,4 +1,4 @@
 #include "HALAL/Models/MPUManager/MPUManager.hpp"
```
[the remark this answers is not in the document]
FoniksFox: Sovled in new commit, now uses a linker symbol instead

STM32H723ZGTX_FLASH.ld
```diff
@@ -179,24 +220,67 @@ SECTIONS
...
+    *(.mpu_ram_d3_nc)
+    *(.mpu_ram_d3_nc*)
   } >RAM_D3
+
+  /* CALCULATE PADDING FOR MPU SUBREGIONS (D3) */
+  _d3_size = SIZEOF(.mpu_ram_d3_nc);
+  /* Find next power of 2 (up to 32KB for 16KB RAM) */
+  _d3_p2 = (_d3_size <= 32) ? 32 : (_d3_size <= 64) ? 64 : (_d3_size <= 128) ? 128 : (_d3_size <= 256) ? 256 : 
+           (_d3_size <= 512) ? 512 : (_d3_size <= 1024) ? 1024 : (_d3_size <= 2048) ? 2048 : 
+           (_d3_size <= 4096) ? 4096 : (_d3_size <= 8192) ? 8192 : (_d3_size <= 16384) ? 16384 : 32768;
```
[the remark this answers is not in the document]
FoniksFox: They're documented in the wiki and in the MPU driver.

Inc/HALAL/Models/MPU.hpp
```diff
@@ -0,0 +1,398 @@
...
+        }
+        // Align totals to 32 bytes, just in case
+        sizes.d1_nc_total = align_up(sizes.d1_nc_total, 32);
+        sizes.d1_c_total  = align_up(sizes.d1_c_total, 32);
+        sizes.d2_nc_total = align_up(sizes.d2_nc_total, 32);
+        sizes.d2_c_total  = align_up(sizes.d2_c_total, 32);
+        sizes.d3_nc_total = align_up(sizes.d3_nc_total, 32);
+        sizes.d3_c_total  = align_up(sizes.d3_c_total, 32);
+        return sizes;
+    }
```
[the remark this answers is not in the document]
FoniksFox: You said it, the linker will catch it, adding more checks is just confusing

Inc/HALAL/Models/MPU.hpp
```diff
@@ -0,0 +1,420 @@
...
+        if constexpr (N == 0) {
+            return {};
+        } else {
+            std::array<Config, N> cfgs{};
+            
+            uint32_t offsets_nc[3] = {}; // D1, D2, D3
+            uint32_t offsets_c[3] = {};  // D1, D2, D3
+            uint32_t assigned_offsets[N]; 
+
+            size_t alignments[] = {32, 16, 8, 4, 2, 1};
```
StefanCostea: make this static private for the class maybe

Inc/HALAL/Models/MPU.hpp
```diff
@@ -0,0 +1,420 @@
...
+         * @param force_cache_alignment If true, forces the buffer to be cache line aligned (32 bytes, takes the rest as padding).
+         */
+        consteval Buffer(MemoryType type = MemoryType::NonCached, MemoryDomain domain = MemoryDomain::D2, bool force_cache_alignment = false)
+        : e{
+            domain,
+            type,
+            force_cache_alignment ? 32 : alignof(T),
+            sizeof(T)}
+        {
+            static_assert(alignof(T) <= 32, "Requested type has alignment greater than cache line size (32 bytes).");
```
StefanCostea: minor:
`
static_assert(alignments.contain(alignof(T)));
`

STM32H723ZGTX_FLASH.ld
```diff
@@ -118,6 +134,39 @@ SECTIONS
...
+    
+    /* Ethernet Rx Pool */
+    . = ALIGN(32);
+    *(.Rx_PoolSection)
+  } >RAM_D1
+
+  /* CALCULATE PADDING FOR MPU SUBREGIONS (D1) */
+  _d1_size = SIZEOF(.mpu_ram_d1_nc);
+  /* Find next power of 2 (up to 512KB for 320KB RAM) */
+  _d1_p2 = (_d1_size <= 32) ? 32 : (_d1_size <= 64) ? 64 : (_d1_size <= 128) ? 128 : (_d1_size <= 256) ? 256 :
```
StefanCostea: Look if our linker supports 
`
_d1_p2 = (1 << LOG2CEIL(MAX(32, _d1_size)));
`

STM32H723ZGTX_RAM.ld
```diff
@@ -161,6 +210,115 @@ SECTIONS
...
+    *(.mpu_ram_d3_nc.user)
+    
+    /* New MPU system buffers */
+    *(.mpu_ram_d3_nc.buffer)
+  } >RAM_D3
+
+  /* CALCULATE PADDING FOR MPU SUBREGIONS (D3) */
+  _d3_size = SIZEOF(.mpu_ram_d3_nc);
+  /* Find next power of 2 */
+  _d3_p2 = (_d3_size <= 32) ? 32 : (_d3_size <= 64) ? 64 : (_d3_size <= 128) ? 128 : (_d3_size <= 256) ? 256 :
```
StefanCostea: Same as above

StefanCostea (changes requested):

Inc/HALAL/Models/MPU.hpp
FoniksFox: Done

Inc/HALAL/Models/MPU.hpp
FoniksFox: Makes sense, done

PR #538, opened 2025-12-14
Feat/MDMA

Cantonplas:
New class to manage the MDMA peripheral. It works on linked list mode. The mdma recieves the first linked list node and will store it in a queue until the next stlib::update where each of the 8 mdma channels will process a request. The nodes need to be stored on a non-cached buffer or shit will go very wrong

To create a linked list node all you have to do is create a instance of the MDMA::LinkedListNode(args, buffer + offset,type_size);

### DIFF
--- a/Inc/HALAL/Models/MDMA/MDMA.hpp
+++ b/Inc/HALAL/Models/MDMA/MDMA.hpp
@@ -4,12 +4,15 @@
 #include "stm32h7xx_hal.h"
 #include "ErrorHandler/ErrorHandler.hpp"
 #include "HALAL/Models/MPUManager/MPUManager.hpp"
+#include "HALAL/Models/MPU.hpp"
 #include <array>
 #include <tuple>
 #include <vector>
 #include <unordered_map>
 #include <type_traits>
 #include <utility>
+
+
 
 #ifdef MDMA
 #undef MDMA
@@ -141,24 +144,26 @@ private:
 };
 
     private:
+    static D1_NC MDMA_LinkNodeTypeDef internal_nodes[8];
+
     struct Instance{
     public:
         MDMA_HandleTypeDef handle;
         uint8_t id;
         volatile bool* done;
-        MDMA_LinkNodeTypeDef transfer_node;
+        MDMA_LinkNodeTypeDef* transfer_node;
 
         Instance()
             : handle{}
             , id(0U)
             , done(nullptr)
-            , transfer_node{}
+            , transfer_node(nullptr)
         {}
 
         Instance(MDMA_HandleTypeDef handle_,
                  uint8_t id_,
                  volatile bool* done_,
-                 MDMA_LinkNodeTypeDef transfer_node_)
+                 MDMA_LinkNodeTypeDef* transfer_node_)
             : handle(handle_)
             , id(id_)
             , done(done_)
--- a/Src/HALAL/Models/MDMA/MDMA.cpp
+++ b/Src/HALAL/Models/MDMA/MDMA.cpp
@@ -1,5 +1,7 @@
 #include "HALAL/Models/MDMA/MDMA.hpp"
 
+D1_NC MDMA_LinkNodeTypeDef MDMA::internal_nodes[8];
+
 #include <algorithm>
 
 std::bitset<8> MDMA::instance_free_map{};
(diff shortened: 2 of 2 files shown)


Inc/HALAL/Models/MDMA/MDMA.hpp
```diff
@@ -0,0 +1,180 @@
...
+        init_node(source_ptr, dest_ptr, size);
+    }
+
+    void set_next(MDMA_LinkNodeTypeDef* next_node) { node.CLAR = reinterpret_cast<uint32_t>(next_node); }
+    void set_destination(void* destination) 
+    { 
+        uint32_t destination_adress = reinterpret_cast<uint32_t>(destination);
+        node.CDAR = destination_adress;
+        node.CTBR &= ~MDMA_CTBR_DBUS;
+        if (destination_adress < 0x00010000)
```
FoniksFox: set_destination should check for domain 2 and 3 as well, only AXI SRAM should have SBUS = 0 (AXI), everything else should have SBUS = 1 (AHB)

FoniksFox: Same for set_source

Inc/HALAL/Models/MDMA/MDMA.hpp
```diff
@@ -0,0 +1,180 @@
...
+    auto get_node() -> MDMA_LinkNodeTypeDef* { return &node; }
+    auto get_size() -> uint32_t { return node.CBNDTR; }
+    auto get_destination() -> uint32_t { return node.CDAR; }
+    auto get_source() -> uint32_t { return node.CSAR; }
+    auto get_next() -> MDMA_LinkNodeTypeDef* { return reinterpret_cast<MDMA_LinkNodeTypeDef*>(node.CLAR); }
+
+private:
+    alignas(8) MDMA_LinkNodeTypeDef node;
+
+    void init_node(void* src, void* dst, size_t size)
```
FoniksFox: Nodes are configured with suboptimal settings, it works but should be noted with a comment at least.
Would be great to have the nodes automatically be configured based on the type of the src

Inc/HALAL/Models/MDMA/MDMA.hpp
```diff
@@ -0,0 +1,180 @@
...
+
+    /**
+     * @brief A method to start a transfer from source to destination using MDMA linked list
+     *  
+     * @param source_address The source address for the transfer.
+     * @param destination_address The destination address for the transfer.
+     * @param data_length The length of data to be transferred.
+     * @param check A reference boolean that will be set to true if the transfer was successfully done, false otherwise.
+     */
+    static void transfer_data(uint8_t* source_address, uint8_t* destination_address, const uint32_t data_length, bool* done=nullptr);
```
FoniksFox: Should use a volatile bool const* to ensure c++ behaves correctly

Src/HALAL/Models/MDMA/MDMA.cpp
```diff
@@ -0,0 +1,296 @@
...
+    const uint32_t clear_flags = MDMA_FLAG_TE | MDMA_FLAG_CTC | MDMA_FLAG_BT | MDMA_FLAG_BFTC | MDMA_FLAG_BRT;
+    __HAL_MDMA_CLEAR_FLAG(&instance.handle, clear_flags);
+
+    __HAL_MDMA_ENABLE_IT(&instance.handle, MDMA_IT_TE | MDMA_IT_CTC);
+
+    __HAL_MDMA_ENABLE(&instance.handle);
+
+    if (HAL_MDMA_GenerateSWRequest(&instance.handle) != HAL_OK)
+    {
+        instance.handle.State = HAL_MDMA_STATE_BUSY;
```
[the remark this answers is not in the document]
FoniksFox: For this maybe you should just return a bool, false is case the hal fails. Because if you set the handle state to busy then you need to manually set it to ready later, I think?

Src/HALAL/Models/MDMA/MDMA.cpp
```diff
@@ -0,0 +1,296 @@
...
+        }
+        if(instance_free_map[i])
+        {
+            Instance& instance = get_instance(i);
+            auto transfer = transfer_queue.top();
+            instance.done = transfer.second;
+            prepare_transfer(instance, transfer.first);
+            transfer_queue.pop();
+        }
+    }
```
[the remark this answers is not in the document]
FoniksFox: instance_free_map[i] is set to false before the transfer is started, so there should be no problem with that

Inc/HALAL/Models/MDMA/MDMA.hpp
FoniksFox: Also should check that the user is not trying to access ITCM or DTCM memory, since no peripheral can access those memories

FoniksFox (dismissed):
LGTM, only needs to change the node to use the new MPU whenever the pr is accepted, everything else seems fine.

PR #543, opened 2025-12-25
fix(GPIO): Alternative Function numbers were inverted

FoniksFox:
Just what the title says + Make inscribes return std::size_t (the result of the add operation) so that they are easier to use from higher domains

### DIFF
--- a/Inc/HALAL/Models/GPIO.hpp
+++ b/Inc/HALAL/Models/GPIO.hpp
@@ -196,8 +196,8 @@ struct GPIODomain {
       }
     }
 
-    template <class Ctx> consteval void inscribe(Ctx &ctx) const {
-      ctx.template add<GPIODomain>(e, this);
+    template <class Ctx> consteval std::size_t inscribe(Ctx &ctx) const {
+      return ctx.template add<GPIODomain>(e, this);
     }
   };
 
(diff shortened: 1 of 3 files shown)


victor-Lopez25 (approved):
👍

jorgesg82 (dismissed):
They are not inverted I think. The AF0 is the LSB. Check the function `valid_af`:
```cpp
inline constexpr bool valid_af(const AlternateFunction af) const {
  if (af == AlternateFunction::NO_AF)
    return true;
  return ((1 << static_cast<uint8_t>(af)) & afs) != 0;
}
```

jorgesg82:
From my point of view, the instantiated pins in Inc/HALAL/Models/Pin.hpp, the AlternateFunction enum, and the valid_af function match one with each other. I think the only thing that does NOT match with the rest is the cast in the build function, so changing that to match with the others will be the less disturbing fix, don't you think?

Otherwise you will have to invert the enum, to modify the valid_af function, and to change all the bitmaps (that would be for sure the most painful thing of them)

FoniksFox:
> From my point of view, the instantiated pins in Inc/HALAL/Models/Pin.hpp, the AlternateFunction enum, and the valid_af function match one with each other. I think the only thing that does NOT match with the rest is the cast in the build function, so changing that to match with the others will be the less disturbing fix, don't you think?
> 
> Otherwise you will have to invert the enum, to modify the valid_af function, and to change all the bitmaps (that would be for sure the most painful thing of them)

Yeah, that's true, will do that, I'll also add a comment to the enum clarifying that

oganigl (approved):
It seems fine to me.

PR #549, opened 2026-01-20
Compile time State Machine

Cantonplas:
Refactor of the state machine to make compile time checks. Now the state machine can only accept states that are part of the same state enum and checks it in compile time.

To create a state machine, first create the states and its transitions with make_state and then create the state machine using this states. Then call the classical functions like add_enter_action, or add_low_precision_cyclic_action like always. Now only func pointers are allowed, no std::function

Wiki entry: https://wiki.hyperloopupv.com/es/firmware/st-lib/StateMachine

State Orders and that stuff i've adapted them to make the code compile, I cant ensure they will work. We dont lose much because previously they didnt work :p

### DIFF
--- a/Inc/HALAL/Models/GPIO.hpp
+++ b/Inc/HALAL/Models/GPIO.hpp
@@ -71,6 +71,7 @@ struct GPIODomain {
       return GPIO_SPEED_FREQ_VERY_HIGH;
     }
   }
+  // Note: AF mapping is inverted: AF0 -> 15, AF1 -> 14, ..., AF15 -> 0, it is staticly casted and inverted later
   enum class AlternateFunction : uint8_t {
     NO_AF = 20,
     AF0 = 15,
@@ -189,10 +190,14 @@ struct GPIODomain {
       if (!pin.valid_af(af)) {
         compile_error("Alternate function not valid for this pin");
       }
+
+      if ((mode == OperationMode::ALT_PP || mode == OperationMode::ALT_OD) && af == AlternateFunction::NO_AF) {
+        compile_error("Alternate function must be specified for alternate modes");
+      }
     }
 
-    template <class Ctx> consteval void inscribe(Ctx &ctx) const {
-      ctx.template add<GPIODomain>(e, this);
+    template <class Ctx> consteval std::size_t inscribe(Ctx &ctx) const {
+      return ctx.template add<GPIODomain>(e, this);
     }
   };
 
@@ -221,7 +226,7 @@ struct GPIODomain {
       GPIO_InitStruct.Pull = to_hal_pull(e.pull);
       GPIO_InitStruct.Speed = to_hal_speed(e.speed);
       if (e.mode == OperationMode::ALT_PP || e.mode == OperationMode::ALT_OD) {
-        GPIO_InitStruct.Alternate = static_cast<uint32_t>(e.af);
+        GPIO_InitStruct.Alternate = 15 - static_cast<uint32_t>(e.af); // AF mapping inversion
       }
 
       cfgs[i].init_data = std::make_tuple(e.port, GPIO_InitStruct);
--- a/Inc/ST-LIB_LOW/DigitalInput2.hpp
+++ b/Inc/ST-LIB_LOW/DigitalInput2.hpp
@@ -18,10 +18,10 @@ struct DigitalInputDomain {
                            GPIODomain::Speed speed = GPIODomain::Speed::Low)
         : gpio{pin, GPIODomain::OperationMode::INPUT, pull, speed} {}
 
-    template <class Ctx> consteval void inscribe(Ctx &ctx) const {
-      const auto gpio_idx = ctx.template add<GPIODomain>(gpio.e, &gpio);
+    template <class Ctx> consteval std::size_t inscribe(Ctx &ctx) const {
+      const auto gpio_idx = gpio.inscribe(ctx);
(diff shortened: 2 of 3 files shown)


Inc/ST-LIB_LOW/StateMachine/StateMachine.hpp
```diff
@@ -17,333 +20,579 @@ using ms = std::chrono::milliseconds;
...
+  StateEnum state;
+  FixedVector<Transition<StateEnum>, NTransitions> transitions;
+
+  public:
+  [[no_unique_address]]FixedVector<uint16_t,Number_of_state_orders> state_orders_ids = {};
+  static constexpr size_t transition_count = NTransitions;
+
+  template <typename... T>
+        requires are_transitions<StateEnum, T...>
+    consteval State(StateEnum state, T... transitions)
```
FoniksFox: Why not pass the transitions as template parameters directly, and infer `NTransitions` from sizeof?

Inc/ST-LIB_LOW/StateMachine/StateMachine.hpp
```diff
@@ -17,333 +20,579 @@ using ms = std::chrono::milliseconds;
...
-  template <class TimeUnit>
-  TimedAction *
-  add_mid_precision_cyclic_action(function<void()> action,
-                                  chrono::duration<int64_t, TimeUnit> period);
+  constexpr void add_state_order(uint16_t id)
+  {
+    state_orders_ids.push_back(id);
+  }
 
   template <class TimeUnit>
```
FoniksFox: Constrain TimeUnit to integral or something similar, whatever can ensure the class always works, so that it doesn't try to accept a type it can't

Inc/ST-LIB_LOW/StateMachine/StateMachine.hpp
```diff
@@ -17,333 +20,579 @@ using ms = std::chrono::milliseconds;
...
+    virtual void set_on(bool is_on) = 0;
+    virtual void force_change_state(size_t state) = 0;
+    virtual size_t get_current_state_id() const = 0;
+    constexpr bool operator==(const IStateMachine&) const = default;
+};
 
-  unordered_map<state_id, State> &get_states();
+template <class StateEnum, size_t NStates, size_t NTransitions>
+class StateMachine : public IStateMachine {
+private:
```
FoniksFox: Same as before, would be better to pass states and transitions in compile-time so that they can be optimized away as much as possible

Inc/ST-LIB_LOW/StateMachine/StateMachine.hpp
```diff
@@ -17,333 +20,579 @@ using ms = std::chrono::milliseconds;
...
+        requires are_states<StateEnum, S...>
+    consteval StateMachine(StateEnum initial_state, S... states) 
+    : current_state(initial_state),
+      states{states...}
+     {        
+        size_t offset = 0;
+        ((process_state(states, offset),
+          offset += states.get_transitions().size()),
+         ...);
+    }
```
FoniksFox: Should maybe use a concept for S

Inc/ST-LIB_LOW/StateMachine/StateMachine.hpp
Cantonplas: Because it's easier to implement and for the user doing it the way its done right now. The transitions are made creating the state with its helper function that constructs the states and the to create the state machine you only have to pass as parameters the states

Inc/ST-LIB_LOW/StateMachine/StateMachine.hpp
```diff
@@ -17,333 +20,579 @@ using ms = std::chrono::milliseconds;
...
+    for(size_t i = 0; i < cyclic_actions.size(); i++)
+    {
+      TimedAction& timed_action = cyclic_actions[i];
+      if(timed_action.action == nullptr){continue;}
+      if(!timed_action.is_on){ continue;}
+
+      switch(timed_action.alarm_precision)
+      {
+      case LOW_PRECISION:
+        Time::unregister_low_precision_alarm(timed_action.id);
```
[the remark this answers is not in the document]
Cantonplas: Done changing it, forgot it existed sorry

Inc/ST-LIB_LOW/StateMachine/StateMachine.hpp
Cantonplas: Its all done in compile-time, only the scheduler calls and transition checks are done in run-time. All the creation of states, the adition of cyclic actions, enter actions, transitions,etc... is done in compile-time

Inc/ST-LIB_LOW/StateMachine/StateMachine.hpp
```diff
@@ -17,333 +21,560 @@ using ms = std::chrono::milliseconds;
...
+
+/// Interface for State Machines to allow other classes to interact with the state machine without knowing its implementation
+class IStateMachine {
+    public:
+    virtual constexpr ~IStateMachine() = default;
+    virtual void check_transitions() = 0;
+    virtual void set_on(bool is_on) = 0;
+    virtual void force_change_state(size_t state) = 0;
+    virtual size_t get_current_state_id() const = 0;
+    constexpr bool operator==(const IStateMachine&) const = default;
```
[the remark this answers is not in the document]
Cantonplas: Si no iba bro :+1:

PR #554, opened 2026-01-27
Refactor/spi

FoniksFox:
Refactor SPI to the new compile infrastructure.

It is a refactor of the base driver, it discards all the logic that was built on top of the driver and mixed with it in the previous implementation.

Uses the new DMA (includes the refactor here).

Allows any desired spi config you may need.

If you want a protocol on top of it, you have to implement it (it is out of the scope for this pr, I'm planning on doing a general protocol class that's a mixin and can use any driver, format and operations). But it is pretty easy to use with packets (will add a better explanation in the wiki later):
- Define all the packets you want to transmit via spi
- Take the biggest packet size as the size of your buffers
- Declare tx and rx buffers
- Write and read packets to these buffers, use the id to know how to interpret the data, since you will allways be sending / receiving the max amount of data
- To handle the case where the slave wants to send to the master, you either have a separate pin for it, or you make the master poll all the time and make the slave know it is a poll because it uses a specific id or something, I don't know, that's on you to decide

It uses a different api for master and slave to make clear the difference in behaviour (master instantly start the operation, while slave prepares it but must wait for the master).

The dma doesn't work right now with the runes as it is, you need to do some shenanigans to make things like adc work, not very compatible right now...

Also changed from static library to object library because things, ask @jorgesg82

### DIFF
--- a/Tests/Time/common_tests.cpp
+++ b/Tests/Time/common_tests.cpp
@@ -5,6 +5,20 @@ std::string ErrorHandlerModel::line;
 std::string ErrorHandlerModel::func;
 std::string ErrorHandlerModel::file;
 
+namespace ST_LIB::TestErrorHandler {
+bool fail_on_error = true;
+int call_count = 0;
+
+void reset() {
+    fail_on_error = true;
+    call_count = 0;
+}
+
+void set_fail_on_error(bool enabled) {
+    fail_on_error = enabled;
+}
+} // namespace ST_LIB::TestErrorHandler
+
 void ErrorHandlerModel::SetMetaData(int line, const char * func, const char * file){
 		ErrorHandlerModel::line = to_string(line);
 		ErrorHandlerModel::func = string(func);
@@ -12,7 +26,11 @@ void ErrorHandlerModel::SetMetaData(int line, const char * func, const char * fi
 }
 
 void ErrorHandlerModel::ErrorHandlerTrigger(string format, ... ){
-    EXPECT_EQ(1, 0);
+    (void)format;
+    ST_LIB::TestErrorHandler::call_count++;
+    if(ST_LIB::TestErrorHandler::fail_on_error){
+        EXPECT_EQ(1, 0);
+    }
 }
 
 void ErrorHandlerModel::ErrorHandlerUpdate(){}
--- a/Tests/dma2_test.cpp
+++ b/Tests/dma2_test.cpp
@@ -7,6 +7,12 @@
 #include "MockedDrivers/NVIC.hpp"
 #include "MockedDrivers/mocked_hal_dma.hpp"
 
+namespace ST_LIB::TestErrorHandler {
+void reset();
+void set_fail_on_error(bool enabled);
+extern int call_count;
+} // namespace ST_LIB::TestErrorHandler
+
 extern "C" {
 void DMA1_Stream0_IRQHandler(void);
 void DMA2_Stream7_IRQHandler(void);
@@ -97,6 +103,30 @@ constexpr std::array<ST_LIB::DMA_Domain::Entry, 2> irq_entries{{
 constexpr auto irq_cfg = ST_LIB::DMA_Domain::build<2>(
     std::span<const ST_LIB::DMA_Domain::Entry, 2>{irq_entries});
 
+constexpr std::array<ST_LIB::DMA_Domain::Entry, 2> i2c_entries{{
+    {.instance = ST_LIB::DMA_Domain::Peripheral::i2c2,
+     .stream = ST_LIB::DMA_Domain::Stream::dma1_stream2,
+     .irqn = DMA1_Stream2_IRQn,
+     .id = 0},
+    {.instance = ST_LIB::DMA_Domain::Peripheral::i2c2,
+     .stream = ST_LIB::DMA_Domain::Stream::dma1_stream3,
+     .irqn = DMA1_Stream3_IRQn,
+     .id = 1},
+}};
+
+constexpr auto i2c_cfg = ST_LIB::DMA_Domain::build<2>(
+    std::span<const ST_LIB::DMA_Domain::Entry, 2>{i2c_entries});
+
+constexpr std::array<ST_LIB::DMA_Domain::Entry, 1> none_entries{{
+    {.instance = ST_LIB::DMA_Domain::Peripheral::none,
+     .stream = ST_LIB::DMA_Domain::Stream::dma1_stream4,
+     .irqn = DMA1_Stream4_IRQn,
+     .id = 0},
+}};
+
+constexpr auto none_cfg = ST_LIB::DMA_Domain::build<1>(
+    std::span<const ST_LIB::DMA_Domain::Entry, 1>{none_entries});
+
 void clear_nvic_enables() {
   for (auto &reg : NVIC->ISER) {
     reg = 0U;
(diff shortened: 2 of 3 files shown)


Inc/HALAL/Models/DMA/DMA2.hpp
```diff
@@ -0,0 +1,401 @@
...
+
+        
+        template <std::size_t N> struct Init {
+            static inline std::array<Instance, N> instances{};
+
+            static void init(std::span<const Config, N> cfgs) {
+                if (N == 0) return;
+                __HAL_RCC_DMA1_CLK_ENABLE();
+	            __HAL_RCC_DMA2_CLK_ENABLE();
+                for (std::size_t i = 0; i < N; ++i) {
```
victor-Lopez25, line 380: Would it be worth it to only enable DMA1/DMA2 not both if only one of them was used?

Inc/HALAL/Models/SPI/SPI2.hpp
```diff
@@ -0,0 +1,1134 @@
...
+using ST_LIB::SPIConfigTypes;
+
+// Forward declaration of IRQ handlers and HAL callbacks
+extern "C" {
+    void SPI1_IRQHandler(void);
+    void SPI2_IRQHandler(void);
+    void SPI3_IRQHandler(void);
+    void SPI4_IRQHandler(void);
+    void SPI5_IRQHandler(void);
+    void SPI6_IRQHandler(void);
```
victor-Lopez25, line 29: If the code that uses SPI2.hpp doesn't need to access the IRQHandlers, these don't need to be defined in SPI2.hpp.

It is a design choice but these could be only in the cpp file since it gets linked in the final executable the same.

Inc/HALAL/Models/SPI/SPI2.hpp
```diff
@@ -0,0 +1,1134 @@
...
+        if (peripheral == SPIPeripheral::spi6) {
+            if (compare_pin(pin, PA4)) return GPIODomain::AlternateFunction::AF8;
+            if (compare_pin(pin, PA5)) return GPIODomain::AlternateFunction::AF8;
+            if (compare_pin(pin, PA6)) return GPIODomain::AlternateFunction::AF8;
+            if (compare_pin(pin, PA7)) return GPIODomain::AlternateFunction::AF8;
+            if (compare_pin(pin, PA15)) return GPIODomain::AlternateFunction::AF7;
+            if (compare_pin(pin, PB3)) return GPIODomain::AlternateFunction::AF8;
+            if (compare_pin(pin, PB4)) return GPIODomain::AlternateFunction::AF8;
+            if (compare_pin(pin, PB5)) return GPIODomain::AlternateFunction::AF8;
+        }
```
victor-Lopez25, line 103: You made an operator== method for pins, reminding you here since you might want to use it.

Inc/HALAL/Models/SPI/SPI2.hpp
```diff
@@ -0,0 +1,1134 @@
...
+                    compile_error("Invalid SCK pin for SPI6");
+                }
+                if (!compare_pin(miso_pin, PB4) &&
+                    !compare_pin(miso_pin, PG12) &&
+                    !compare_pin(miso_pin, PA6)) {
+                    compile_error("Invalid MISO pin for SPI6");
+                }
+                if (!compare_pin(mosi_pin, PB5) &&
+                    !compare_pin(mosi_pin, PG14) &&
+                    !compare_pin(mosi_pin, PA7)) {
```
victor-Lopez25, line 361: Reminding you here for the operator==() method you made to compare pins in case you want to use it here

Inc/HALAL/Models/SPI/SPI2.hpp
```diff
@@ -0,0 +1,1134 @@
...
+        template <typename E, size_t S>
+        bool send(span<E, S> data) {
+            if (data.size_bytes() % frame_size != 0) {
+                ErrorHandler("SPI data size (%d) not aligned to frame size (%d)", data.size_bytes(), frame_size);
+                return false;
+            }
+            auto error_code = HAL_SPI_Transmit(&spi_instance.hspi, (uint8_t*)data.data(), data.size_bytes() / frame_size, 10);
+            return check_error_code(error_code);
+        }
+
```
victor-Lopez25, line 508: Span? 😭

Inc/HALAL/Models/SPI/SPI2.hpp
```diff
@@ -0,0 +1,1134 @@
...
+        template <typename T>
+        bool receive_DMA(T& data, volatile bool* operation_flag = nullptr) requires std::is_trivially_copyable_v<T> {
+            spi_instance.operation_flag = operation_flag;
+             if (sizeof(T) % frame_size != 0) {
+                ErrorHandler("SPI data size (%d) not aligned to frame size (%d)", sizeof(T), frame_size);
+                return false;
+            }
+            auto error_code = HAL_SPI_Receive_DMA(&spi_instance.hspi, reinterpret_cast<uint8_t*>(&data), sizeof(T) / frame_size);
+            return check_error_code(error_code);
+        }
```
victor-Lopez25, line 658: Nitpick: you can avoid checking if operation_flag is nullptr or not by pointing to some variable you have defined in SPI2.cpp by default

Src/HALAL/Services/ADC/ADC.cpp
```diff
@@ -68,7 +68,7 @@ void ADC::turn_on(uint8_t id){
 
 	uint32_t buffer_length = peripheral->init_data.channels.size();
 	if (HAL_ADC_Start_DMA(peripheral->handle, (uint32_t*) peripheral->dma_data_buffer, buffer_length) != HAL_OK) {
-		ErrorHandler("DMA - %d - of ADC - %d - did not start correctly", peripheral->init_data.dma_stream, id);
+		ErrorHandler("DMA - %d - of ADC - %d - did not start correctly", peripheral->init_data.dma_instance, id);
```
victor-Lopez25, line 71: Nitpick: dma_instance is a pointer so this should be %p instead of %d

Inc/HALAL/Models/SPI/SPI2.hpp
FoniksFox: The handlers are declared as friends in the Instance class so that they can access private data

Inc/HALAL/Models/SPI/SPI2.hpp
FoniksFox: Yes, I did this before adding the method, it is the same though

Inc/HALAL/Models/SPI/SPI2.hpp
FoniksFox: You can use whatever you want, that's why it is templated, span is just a specialization

Inc/HALAL/Models/SPI/SPI2.hpp
FoniksFox: I think nullptr is more explicit, and just adds an if. Avoiding the hal is what would trully make a difference

Inc/HALAL/Models/SPI/SPI2.hpp
victor-Lopez25: Fair enough

Inc/HALAL/Models/DMA/DMA2.hpp
FoniksFox: Not worth it

Inc/HALAL/Models/SPI/SPI2.hpp
FoniksFox: Same as before

Src/HALAL/Services/ADC/ADC.cpp
FoniksFox: Doesn't matter, says Jorge

PR #557, opened 2026-02-02
Feat/new encoder

victor-Lopez25:
This _should_ work for encoder, not the best abstraction, but kind of works for now, will change when changing the abstraction over pwm, dualpwm and inputcapture.
@Cantonplas


Cantonplas (approved):
Durisimo

PR #560, opened 2026-02-03
Hotfix! fixed a bug with a strange case when getting an overflow

victor-Lopez25:
Hotfix! fixed a bug with a strange case when getting an overflow in global_timer_us_ (overflow of 32 bits) when registering multiple tasks just before.


Cantonplas (approved):

oganigl (approved):
Looks fine

PR #566, opened 2026-02-07
Fix/scheduler

victor-Lopez25:
Fixes CNT going over ARR value in scheduler and scheduler not waiting an accurate amount of time.


victor-Lopez25:
Tests won't work most likely because EGR UG (update generation) register value is not simulated. Will fix tomorrow.

victor-Lopez25:
Tests I thought would work didn't and tests that I thought wouldn't work did huh

FoniksFox (dismissed):

jdmarmen (dismissed):

Cantonplas (approved):

FoniksFox (approved):

PR #568, opened 2026-02-09
fix(PWM): Incorrect calculations for duty cycle

FoniksFox:
Just what the title says, pwm didn't work, now it does, proved with a board


victor-Lopez25 (approved):
I fixed this and I think I did ok, tests would be better to check stuff but I have no time

Cantonplas (approved):

PR #584, opened 2026-03-04
Prescaler was not initialized in init

Cantonplas:


victor-Lopez25 (dismissed):
I made this change and I think I did very good

PR #587, opened 2026-03-09
Max out ARR for encoder

victor-Lopez25:
NOTE: something should probably be changed for underflow...

NOPE apparently value starts in the middle of the range so it won't underflow/overflow

note that this change is included in feat/newInputCapture too since I had the change locally :P


FoniksFox (approved):
LGTM

jdmarmen (approved):

PR #590, opened 2026-03-13
ADC implemented with DMA

jorgesg82:
For that purpose, I had to refactor roughly the DMA to distinguish between peripherals that need IRQs and peripherals that are using DMA in circular mode to just move data all the time. So far, only the ADC has been selected as one of those.

I also renamed DMA_Domain to DMADomain.

I also expanded the test coverage for DMA.

The old ADC has been finally removed.

I also refactored all the sensors that use the ADC to align with this new one.


Inc/HALAL/Models/DMA/DMA2.hpp
```diff
@@ -451,25 +456,41 @@ struct DMA_Domain {
...
             __HAL_RCC_DMA2_CLK_ENABLE();
+
             for (std::size_t i = 0; i < N; ++i) {
                 const auto& e = cfgs[i];
-                auto [instance, dma_init, stream, irqn, id] = e.init_data;
+                auto [instance, dma_init, stream, irqn, id, nvic_enabled] = e.init_data;
+                (void)instance;
+                (void)id;
+
+                if (stream == Stream::none) {
```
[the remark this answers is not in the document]
FoniksFox: Init::init is only called once.

Inc/MockedDrivers/mocked_hal_adc.hpp
```diff
@@ -2,18 +2,52 @@
 
 #include "hal_wrapper.h"
 
+#include <cstddef>
```
victor-Lopez25: `stddef.h` is already included in `hal_wrapper.h` by `stm32h7xx_hal_mock.h`

Inc/MockedDrivers/mocked_hal_dma.hpp
```diff
@@ -3,18 +3,21 @@
 #include "hal_wrapper.h"
 
 #include <cstddef>
+#include <cstdint>
```
victor-Lopez25: both of these are already included in `hal_wrapper.h` in `stm32h7xx_hal_mock.h` as `stddef.h` and `stdint.h` (they are the same)

Inc/ST-LIB_LOW/Sensors/NTC/NTC.hpp
```diff
@@ -9,16 +9,22 @@
 This NTC class is not generic. It is only for 10k Ohm, 1976Beta value NTCs.
 */
 #pragma once
-#include <cstdint>
+#include <cstddef>
```
victor-Lopez25: `cstddef` is already included in `ADCSensor.hpp`

victor-Lopez25 (dismissed):
Apart from some unnecessary `#includes` it looks good

PR #612, opened 2026-03-22
Fix issue Bad flash memory Usage

oganigl:
Created a flash size in sector 6 to keep other things differents than code, right now only metadata and Hardfault_log


Cantonplas (dismissed):
LGTM

FoniksFox (dismissed):
LGTM

PR #628, opened 2026-04-22
[ISSUE 617] ProtectionManager Without Ethernet / Fault Runtime Redesign

### LINKED ISSUE #617: ProtectionManager working without Ethernet

jorgesg82:
## Summary

Closes #617

This PR removes the protection subsystem's dependency on Ethernet and redesigns protections, diagnostics, and global fault handling around a compile-time `Board` integration model.

Protections are now declared as compile-time board request objects and are evaluated through the board-specific `Board::ProtectionEngine`. The global fault runtime remains owned by `FaultController`, with fixed top-level states: `OPERATIONAL` and `FAULT`.

## Main Changes

- Replace the old `ProtectionManager` model with a compile-time `ProtectionEngine`.
- Declare protections with `Protections::protection<"name", source>(...)`.
- Pass protection declarations directly to `Board<Policy, ...>` alongside hardware requests.
- Expose protection evaluation through `Board::evaluate_protections()`.
- Remove runtime protection registration and mutable protection registry behavior.
- Remove the low/high protection split and the old boundary-driven integration model.
- Make `FaultController` the owner of the global fault runtime.
- Move real bootstrap wiring to `Board::init()`.
- Install diagnostics sinks and the fault runtime early during bootstrap so early faults are latched and not lost.
- Unify runtime reporting through diagnostics with `PANIC(...)`, `FAULT(...)`, `WARNING(...)`, and `INFO(...)`.
- Separate `FaultCause` from `DiagnosticRecord`.
- Make protection time accumulation use `Scheduler::get_global_tick()` instead of loop-frequency assumptions.
- Add compile-time coverage for the `Board + Protections` contract.
- Update tests, changeset, and documentation to match the new integration model.

## Breaking Changes

- `Board` now requires the fault policy type as its first template argument.
- Protections are no longer registered at runtime with `ProtectionEngine::create_protection(...)`.
- Protection rules are no longer added dynamically with `ProtectionHandle::add_rule(...)`.
- Protection declarations must be compile-time board request objects.
- The user no longer programs transitions to the global `FAULT` state.
- The user no longer owns the top-level runtime state machine.
- The intended bootstrap path is `Board::init()`, not legacy `STLIB::*` or `STLIB_LOW/HIGH::*` entry points.

## New Board API

```cpp
using AppFaultPolicy = ST_LIB::FaultPolicy<app_machine, on_fault_enter>;
using AppBoard = ST_LIB::Board<AppFaultPolicy, led, pwm, adc>;
```

If the application does not need a nested machine:

```cpp
using AppBoard = ST_LIB::Board<ST_LIB::FaultPolicyNoMachine<on_fault_enter>, led, pwm, adc>;
```

If the application needs neither a nested machine nor a fault-entry callback:

```cpp
using AppBoard = ST_LIB::Board<ST_LIB::DefaultFaultPolicy, led, pwm, adc>;
```

## New Protection API

```cpp
float bus_voltage = 0.0f;

inline constexpr auto bus_voltage_protection =
    Protections::protection<"bus_voltage", bus_voltage>(
        Protections::Rules::below(350.0f, 370.0f),
        Protections::Rules::time_accumulation(20.0f, 15.0f, 0.5f)
    );

using AppBoard = ST_LIB::Board<
    ST_LIB::DefaultFaultPolicy,
    bus_voltage_protection,
    led,
    adc>;
```

## Runtime Loop

```cpp
AppBoard::init();

while (1) {
    FaultController::check_transitions();
    AppBoard::evaluate_protections();
    Diagnostics::Hub::flush();
}
```

## Fault Entry Contract

- Protections enter global fault through `FaultController::request_fault(...)` internally.
- `PANIC(...)` and `FAULT(...)` also map to `FaultController::request_fault(...)`.
- `WARNING(...)` and `INFO(...)` publish diagnostics only and do not enter global fault.
- User code should prefer `PANIC(...)` and `FAULT(...)` over direct `FaultController::request_fault(...)` usage.

## Migration Guide

1. Replace legacy board declarations with `Board<YourFaultPolicy, ...>`.
2. Move any operational user machine under `FaultPolicy<app_machine, on_fault_enter>`.
3. Stop programming direct transitions to global `FAULT`.
4. Replace old protection registration with compile-time `Protections::protection<"name", source>(...)` declarations.
5. Pass protection declarations to `Board<Policy, ...>`.
6. Replace old reporter usage with `PANIC(...)`, `FAULT(...)`, `WARNING(...)`, and `INFO(...)`.
7. Ensure bootstrap happens through `Board::init()`.
8. Drive protections through `Board::evaluate_protections()` in the main loop.

## Validation

- Simulator build passes.
- Simulator test suite passes.
- `board-debug` build passes.
- Compile-time `Board + Protections` contract check passes.
- Repeated `time_accumulation` evaluation test passes.


Inc/HALAL/Services/PWM/PhasedPWM/PhasedPWM.hpp.old
victor-Lopez25, line 1: `PhasedPWM.old` has no 'new' counterpart, others that were deleted already have a new implementation, like `Encoder` or `InputCapture`.

Maybe it would be fine to keep the old implementations that don't have a new implementation done

Inc/ST-LIB_LOW/HalfBridge/HalfBridge.hpp.old
victor-Lopez25, line 1: Same as the other, this is not implemented and might be good to keep until it is implemented.

Src/HALAL/Models/TimerDomain/TimerDomain.cpp
```diff
@@ -31,7 +31,7 @@ TimerDomain::InputCaptureInfo input_capture_info_dummy = {
 
 TimerDomain::InputCaptureInfo* TimerDomain::input_capture_info[max_instances]
                                                               [input_capture_channels] = {
-                                                                  &input_capture_info_dummy
+                                                                  {&input_capture_info_dummy}
```
victor-Lopez25, line 34: What's this for?

Src/HALAL/Services/Diagnostics/DiagnosticFormatter.cpp
```diff
@@ -0,0 +1,267 @@
...
+    if (offset >= buffer_size - 1) {
+        buffer[buffer_size - 1] = '\0';
+        return;
+    }
+
+    va_list arguments;
+    va_start(arguments, format);
+    vsnprintf(buffer + offset, buffer_size - offset, format, arguments);
+    va_end(arguments);
+}
```
victor-Lopez25: It would be better to have a struct, like a string builder to keep the current offset since with this it will call `bounded_strnlen` every call which is O(n) and pretty slow.

For an example implementation of a string builder, see: https://github.com/victor-Lopez25/viclib/blob/main/vl_build.h#L1099

Src/HALAL/Services/Time/Scheduler.cpp
```diff
@@ -99,9 +102,12 @@ void Scheduler::update() {
 
 uint64_t Scheduler::get_global_tick() {
     SchedLock();
-    uint64_t val = global_tick_us_ + Scheduler_global_timer->CNT;
+    uint64_t tick = global_tick_us_;
+    if (Scheduler_global_timer != nullptr) {
+        tick += Scheduler_global_timer->CNT;
+    }
```
victor-Lopez25, line 109: `Scheduler::get_global_tick()` assumes the board has been started up so `Scheduler_global_timer` is not null, this condition is not necessary and will make this function slower for no reason

Inc/HALAL/Services/PWM/PhasedPWM/PhasedPWM.hpp.old
FoniksFox: Isn't it just a configuration layer over a PWM? If so, it could just be replaced by a funciton in the pwm class, but I don't actually know how it works so don't take my word

Inc/ST-LIB_LOW/HalfBridge/HalfBridge.hpp.old
FoniksFox: It is just 2 PWMs, there's not much logic in it. In any case I would just refactor it now rather than keep it, so that we can remove the rest of the old code. But it is not so useful really, and I think no one is using it right now

Inc/HALAL/Services/PWM/PhasedPWM/PhasedPWM.hpp.old
victor-Lopez25: It does some different things from a normal pwm and does it all with lower level handling than what was there in the old PWM.cpp and DualPWM.cpp, though there is also a couple comments that imply it was unfinished...

Inc/ST-LIB_LOW/HalfBridge/HalfBridge.hpp.old
victor-Lopez25: It looks like HalfBridge uses Phased PWMs, I agree that this implementation is not worth keeping

Src/HALAL/Models/TimerDomain/TimerDomain.cpp
jorgesg82: To avoid a warning when compiling

victor-Lopez25: I wasn't getting any warnings when compiling that :thinking_face:

Src/HALAL/Services/Diagnostics/DiagnosticFormatter.cpp
```diff
@@ -0,0 +1,256 @@
...
+        if (written < 0) {
+            return;
+        }
+
+        const size_t remaining = buffer_size - offset;
+        if (static_cast<size_t>(written) >= remaining) {
+            offset = buffer_size - 1;
+            buffer[offset] = '\0';
+            return;
+        }
```
victor-Lopez25, line 37: Nitpick: `snprintf` and `vsnprintf` always add a null terminator so it's not necessary to add `buffer[offset] = '\0'` in this case

Src/HALAL/Services/Diagnostics/DiagnosticFormatter.cpp
```diff
@@ -0,0 +1,256 @@
...
+    size_t buffer_size
+) {
+    if (buffer_size == 0) {
+        return;
+    }
+
+    switch (encoding) {
+    case Protections::SampleEncoding::BOOL:
+        snprintf(buffer, buffer_size, "%s", value.bool_value ? "true" : "false");
+        return;
```
victor-Lopez25, line 105: If you know the values are going to be "true" or "false" there's no reason to add this with `snprintf` you can just memcpy (check bounds)

Src/HALAL/Services/Diagnostics/DiagnosticFormatter.cpp
```diff
@@ -0,0 +1,256 @@
...
+            static_cast<unsigned long long>(value.unsigned_value)
+        );
+        return;
+    case Protections::SampleEncoding::FLOAT32:
+        snprintf(buffer, buffer_size, "%.6f", static_cast<double>(value.float32_value));
+        return;
+    case Protections::SampleEncoding::FLOAT64:
+        snprintf(buffer, buffer_size, "%.6f", value.float64_value);
+        return;
+    }
```
victor-Lopez25, line 131: none of the `snprintf` calls are checked, do we always have buffer size for what we are formatting?

Src/ST-LIB_HIGH/Protections/FaultController.cpp
```diff
@@ -0,0 +1,212 @@
...
+    if (src == nullptr) {
+        return 0;
+    }
+
+    size_t length = 0;
+    while (length < max_length && src[length] != '\0') {
+        length++;
+    }
+    return length;
+}
```
victor-Lopez25: Is there a reason to use this implementation instead of stdc's `strnlen`?
[strnlen](https://man7.org/linux/man-pages/man3/strnlen.3.html)

victor-Lopez25 (changes requested):
Looks good, you should check the scheduler comment and the diagnosticformatter comments I added.

Also if you deleted the *.old files, which is fine, you can also remove the directories that now contain no files (if there are any).

Inc/ST-LIB_LOW/ErrorHandler/ErrorHandler.hpp
```diff
@@ -1,65 +1,31 @@
...
 
-    friend class BoundaryInterface;
+class FaultReporter {
+public:
+    static void Trigger(const std::source_location& location, const char* format, ...);
+    static void Flush();
 };
 
-#define ErrorHandler(x, ...)                                                                       \
+#define PANIC(x, ...)                                                                              \
```
StefanCostea, line 23: Deberiais tener algun nvic lock para esto, lo usais por todos lados, si el main loop esta haciendo un panic y de repente salta un interrupt que tambien hace un panic por lo que sea al escribir en pending_records hay una race condition

StefanCostea: Y si tienes una race condition al hacer panic y por lo que sea hard fault, no propagas

StefanCostea: O sabe dios que otro unintended behaviour

Src/HALAL/Services/Diagnostics/DiagnosticSinks.cpp
```diff
@@ -0,0 +1,215 @@
...
+#ifdef STLIB_ETH
+#include "HALAL/Models/Packets/Order.hpp"
+#endif
+
+namespace Diagnostics {
+
+namespace {
+
+constexpr uint16_t DIAGNOSTIC_TCP_ORDER_ID = 3555;
+constexpr size_t transport_buffer_capacity = 2 + 1 + 1 + 2 + 1 + 1 + 1 + 1 + 1 + 2 + 8 +
```
StefanCostea: Que es esto xDDDD

StefanCostea (changes requested):

Inc/ST-LIB_LOW/ErrorHandler/ErrorHandler.hpp
victor-Lopez25: Yo creo q simplemente no se debería de hacer panic en una interrupción

Src/HALAL/Services/Diagnostics/DiagnosticSinks.cpp
victor-Lopez25: JAJAJAJAJJ es el total de todos los sizeof de las cosas que le pone, mira la siguiente vez en el fichero donde está, sería mejor tenerlo con un struct y hacerle sizeof al struct y ya XD

Inc/ST-LIB_HIGH/Protections/Rules.hpp
```diff
@@ -0,0 +1,248 @@
...
+};
+
+template <typename T> struct TimeAccumulationRuleConfig {
+    T fault_threshold{};
+    optional<T> warning_threshold{};
+    float time_window_s{0.0f};
+};
+
+template <ProtectionSample T>
+using RuleDefinition = variant<
```
StefanCostea, line 41: Watchout with this being a variant as evaluating it might need runtime std::visit calls that are costly

StefanCostea (changes requested):

FoniksFox (changes requested):
The default ethernet sink doesn't work properly (working on it)

Inc/ST-LIB_LOW/ErrorHandler/ErrorHandler.hpp
FoniksFox: Es que si no haces panic en una interrupción entonces cuando te llega una interrupción de error qué haces? xd

victor-Lopez25: En qué caso tienes una 'interrupción de error'?
Yo he visto interrupción de fallo pero no sería equivalente a un panic, sería más como añadir uno al contador de fallos e ignorarlo en el main si la frecuencia de fallos es baja.

PR #635, opened 2026-04-28
fix pwm config for channel 4

victor-Lopez25:
(thanks @FoniksFox)


FoniksFox (approved):
I requested this change, I can confirm that it works now

oganigl (approved):
THANKS FONIKS

PR #637, opened 2026-04-29
Hotfix/MPU Part 3

FoniksFox:
The MPU dynamic regions (Non-Cached regions) were being configured with linker symbol values instead of addresses, which caused undefined behaviour that even could acouse hardfault.


jorgecanut (approved):
Worked for me

oganigl (approved):

oganigl:
I've seen it work in live.

PR #639, opened 2026-04-30
Fix/volatile

### LINKED ISSUE #638: Undefined Behaviour due to lack of volatile
Many domains and classes don't use volatile to access non-cached memory, which may cause undefined behavior if the compiler decides to keep the read / write in registers (mainly in release builds).

FoniksFox:
Closes #638 
Fixes lots of small errors in the library related to not using volatiles to access memory that is / should be non-cached.
Check the issue for more information.

There may be more hidden errors elsewhere in the codebase, but this takes care of a good chunk of them and should not let compile any code that doesn't follow the volatile rules (as long as it follows standard practices)


oganigl (approved):

victor-Lopez25 (approved):
LGTM

PR #644, opened 2026-05-08
[643] Fixed new protections don't accept volatile

### LINKED ISSUE #643: New protections don't accept volatile

jorgesg82:
# Volatile Protection Sources

Related PR: [#628 - ProtectionManager Without Ethernet / Fault Runtime Redesign](https://github.com/Hyperloop-UPV/ST-LIB/pull/628)

## Summary

This follow-up extends the compile-time protection model introduced in PR #628 so protections
can use `volatile` variables directly as sample sources.

Protection source storage now preserves cv-qualification through a reference source wrapper,
while rule evaluation still works with plain sample values. This keeps the public protection
declaration API unchanged and avoids making all protection reads volatile by default.

## Main Changes

- Add `ReferenceSampleSource<Storage>` to preserve the storage type used by a protection source.
- Keep `SampleSource<Storage>` as an alias for the reference source wrapper.
- Make `Protection` store the actual readable source type instead of forcing every source into
  `SampleSource<T>`.
- Make `ProtectionEngine` wrap raw variables with a source type that preserves `volatile`.
- Add simulator coverage proving a protection declared over a `volatile float` reads and reports
  the observed value correctly.

## Example

```cpp
inline volatile float temperature_c = 0.0f;

inline constexpr auto temperature_protection =
    Protections::protection<"temperature", temperature_c>(
        Protections::Rules::above(80.0f, 60.0f)
    );
```


FoniksFox (approved):
LGTM and it seems to be working correctly

PR #661, opened 2026-06-12
one parenthesis, 2 hours of my time

oganigl:
Is just one parenthesis


FoniksFox (changes requested):
Ugly. This is not intuitive.
Also, it breaks the API withou much reason.

PR #667, opened 2026-06-21
Feat/clock domain

### LINKED ISSUE #665: Flash latency set to 3 instead of 4
Issue on the clock tree, the hal silently fixes it so it doesn't actually matter, but it is an issue nonetheles

FoniksFox:
Closes #665 

### Add ClockDomain, a centralised clock validator and applicator

Clock configuration is a precomputed `ClockTree` (from a host tool or hand-written). The firmware validates it at compile time and applies it at runtime.  No solver runs in the firmware.

- `ClockTree` stores only decisions (M/N/P/Q/R, prescalers, source enums); all frequencies are derived via accessors (`sysclk(t)`, `source_frequency(t, src)`, etc.)
- `Board<>` takes a mandatory `ClockTree` as the second template parameter
- Peripherals register clock requirements via `ClockDomain::Device` with typed models: `SPIClockModel<Group, MaxBaud, MinBaud>`, `ADCClockModel<MaxADCCLK>`, `SDClockModel<MaxFreq, MinFreq>`
- `ClockDomain::build()` validates PLL ranges, bus consistency, and peripheral requirements at compile time - refuses to compile on failure
- `ClockDomain::Init::init(tree)` applies the validated tree to HAL registers
- ADC prescaler computed from kernel clock at init (no longer user-specified)
- SPI/SD use mandatory frequency ranges
- Removed `HALconfig`

Clock-tree solver was not implement in compile time because of the added complexity (worst cases were not possible to compile unless willing to let the compiler do billions of operations). The solver may be implement at a further date via a brute force function in the domain that only exists on simulator builds, such that you can build your project with the simulation preset, solve the clock-tree and print it to a header file (it could be added as a build step or as a standalone script).


Inc/HALAL/Models/TimerDomain/TimerDomain.hpp
victor-Lopez25, line 1: There is a lot of code duplicated here which shouldn't be, TimerDomain is not to be used like TimerWrapper. TimerWrapper already handles clocks, so remove TimerRequest from `TimerDomain::Config` and `clock_frequency` from the timer instance, as you can already get that from the timeridx or the `TIM_TypeDef`, it's fine to have a runtime check for if a timer is in apb1 or apb2 _if_ you only have the `TIM_TypeDef` and that's exactly the reason why I have that function in TimerDomain.

victor-Lopez25, line 1: If you need the ClockDomain from the TimerDomain to know what apb1 and apb2 frequencies are, you should probably use a global ClockDomain, that would keep it simpler and you could even get the function which uses HAL code to check a `TIM_TypeDef`'s frequency to work. The only place where it is used is in InputCapture, which is pretty important...

Inc/HALAL/Services/Time/TimerWrapper.hpp
victor-Lopez25, line 1: This should just use a global:
```cpp
if constexpr (this->is_on_APB1) {
  return ClockDomain::timer_apb1(global_clock_tree);
} else {
  return ClockDomain::timer_apb2(global_clock_tree);
}
```

You're also adding a PANIC where there's no reason to have one, a timer is always in apb1 or apb2 so it doesn't make much sense to add a possibility for failiure.

Inc/ST-LIB.hpp
```diff
@@ -242,7 +244,11 @@ template <BoardFaultPolicy FaultPolicyT, auto&... devs> struct Board {
             // ...
         };
 
+        // Build: validate the tree against peripheral requirements
+        ClockDomain::build(ctx.template span<ClockDomain>(), ClockTreeDef);
+
         return ConfigBundle{
+            .clock_tree = ClockTreeDef,
             .mpu_cfgs = MPUDomain::template build<mpuN>(ctx.template span<MPUDomain>()),
```
victor-Lopez25, line 252: Is there no way to have it work like the others, so we can simplify creating `Board`?

Inc/ST-LIB.hpp
victor-Lopez25, line 1: If we already have a global ClockDomain instance, `Board::clock_tree()`, why are we passing the clock tree to every domain? I get that you can't access Board from a domain but you could just make it global from ClockDomain since you already know there will only ever be one.

victor-Lopez25 (changes requested):

Inc/ST-LIB.hpp
FoniksFox: Not really, not in any way that makes sense at least. If you have any idea, I'd like to hear it!

PR #672, opened 2026-07-07
fix: Ensure Scheduler::get_global_tick() is monotonic

FoniksFox:
What the title says


victor-Lopez25 (approved):
I'd like a better solution but this seems OK.

jdmarmen (approved):

PR #673, opened 2026-07-08
Feat/visual watchdog

FoniksFox:
Makes the watchdog reset check block the system and visually indicate this state (infinite while with led toggling).

Uses the same leds as hardfault but with a different pattern to visually differentiate them.


jdmarmen (approved):

victor-Lopez25 (approved):
Pretty lights 😁

PR #676, opened 2026-07-12
feat: Add fault propagation to FaultController

jorgesg82:
## Summary

Add automatic fault propagation to `FaultController` so that when a board enters FAULT, it sends the FAULT order (ID 0) to all registered TCP peers, and conversely detects incoming FAULT orders from peers to enter FAULT automatically.

### `FaultController.hpp`
- Added `#ifdef STLIB_ETH` guarded includes for `Order.hpp` and `OrderProtocol.hpp`
- New public API: `register_fault_propagation(OrderProtocol* socket, Order* fault_order)`
- New private members: `FaultPropagationTarget` struct, static storage for up to 8 targets, `propagate_fault()`, `on_fault_order_received()`

### `FaultController.cpp`
- `register_fault_propagation()`: stores socket+order pair and **replaces the order's callback** with `on_fault_order_received` so that when the FAULT order arrives on that socket, FaultController enters FAULT automatically.
- `on_fault_order_received()`: calls `request_fault()` — this is the reception path.
- `propagate_fault()`: iterates all registered sockets and sends the FAULT order via `send_order()`.
- `request_fault()` now calls `propagate_fault()` inside the `if (!faulted)` block — so propagation only fires on the initial fault entry, not on subsequent duplicate calls.

## Loop Prevention

The `if (!faulted)` guard in `request_fault()` is the **only** loop breaker. Each board propagates exactly once, on the `!faulted → faulted` transition. No `fault_received_from_peer` flag needed.

Designed for star topology (VCU at center):

| Scenario | Sequence |
|---|---|
| **Local fault** (e.g. HVBMS faults) | HVBMS: `!faulted → faulted` → propagate to VCU. VCU: `!faulted → faulted` → propagate to CS, HVBMS, PCU, LCU. All replies: already faulted → stop. |
| **CS commands FAULT to VCU** | VCU: `!faulted → faulted` → propagate to CS, HVBMS, PCU, LCU. All replies: already faulted → stop. |

- Works with both `Socket*` and `ServerSocket*` (both inherit from `OrderProtocol`).
- No-op on boards built without `STLIB_ETH` (all code guarded).

## Usage (application side)

```cpp
// After creating sockets and the FAULT order:
FaultController::register_fault_propagation(socket_a, fault_order);
FaultController::register_fault_propagation(socket_b, fault_order);
// ...
```

Registration enables both reception (callback hijack) and sending (propagation on local fault). Without registration, neither works.

### DIFF
--- a/.changesets/fault-propagation.md
+++ b/.changesets/fault-propagation.md
@@ -1,4 +1,4 @@
 release: minor
-summary: Add fault propagation to FaultController for sending FAULT order to remote peers and suppressing re-propagation on received faults
+summary: Add fault propagation to FaultController for sending FAULT order to remote peers with single-shot loop prevention via the `!faulted` guard
 
 Star-topology design: the `!faulted` guard in `request_fault()` is the only loop breaker. Each board propagates exactly once on the `!faulted → faulted` transition, so no `fault_received_from_peer` flag is needed. Works for any node faulting first (local fault or CS-commanded fault).
--- a/.changesets/fault-propagation.md
+++ b/.changesets/fault-propagation.md
@@ -1,0 +1,4 @@
+release: minor
+summary: Add fault propagation to FaultController for sending FAULT order to remote peers and suppressing re-propagation on received faults
+
+Star-topology design: the `!faulted` guard in `request_fault()` is the only loop breaker. Each board propagates exactly once on the `!faulted → faulted` transition, so no `fault_received_from_peer` flag is needed. Works for any node faulting first (local fault or CS-commanded fault).
--- a/Inc/ST-LIB_HIGH/Protections/FaultController.hpp
+++ b/Inc/ST-LIB_HIGH/Protections/FaultController.hpp
@@ -4,6 +4,11 @@
 #include "ST-LIB_HIGH/Protections/ProtectionTypes.hpp"
 #include "StateMachine/StateMachine.hpp"
 
+#ifdef STLIB_ETH
+#include "HALAL/Models/Packets/Order.hpp"
+#include "HALAL/Models/Packets/OrderProtocol.hpp"
+#endif
+
 namespace ST_LIB::TestAccess {
 struct FaultController;
 }
@@ -84,12 +89,20 @@ public:
         reconstruct_runtime_machine<Policy>(
             preserve_preinstalled_fault ? RuntimeState::FAULT : RuntimeState::OPERATIONAL
         );
+#ifdef STLIB_ETH
+        propagation_targets = {};
+        propagation_target_count = 0;
+#endif
     }
 
     static void start();
     static void check_transitions();
     static bool is_faulted();
     static const FaultCause* latched_fault_cause();
+
+#ifdef STLIB_ETH
+    static void register_fault_propagation(OrderProtocol* socket, Order* fault_order);
+#endif
 
 private:
     friend class PanicReporter;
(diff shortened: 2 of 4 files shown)


jorgesg82:
## Copilot Review Responses

Addressed 4 of 7 comments in `6a61145e`:

| # | Comment | Action |
|---|---------|--------|
| 1+5 | Propagation state never reset in `install_runtime()`/`clear()` | **Fixed** — `reset_runtime_storage()` now clears `propagation_targets` and `propagation_target_count` under `STLIB_ETH`. Both `install_runtime<>` and `TestAccess::FaultController::clear()` call `reset_runtime_storage()`, so both paths are covered. |
| 3 | Hard-coded diagnostic metadata (`line=0`, `file=""`) | **Fixed** — `on_fault_order_received()` now uses `__LINE__`, `__func__`, `__FILE__`. |
| 4 | `propagate_fault()` doesn't null-check `fault_order` | **Fixed** — added `&& target.fault_order != nullptr` guard. |
| 6 | Changeset missing blank line after `summary:` | **Fixed** — added blank line + extra context per template. |
| 2 | `register_fault_propagation` doesn't validate order ID == 0 | **Won't fix** — the API is intentionally generic. The application controls which `Order*` it registers; coupling ST-LIB to a specific order ID would be a layering violation. |
| 7 | No unit tests for STLIB_ETH paths | **Won't fix in this PR** — the test suite doesn't build with `STLIB_ETH` defined, so testing these paths requires test infrastructure changes outside this PR's scope. |

jorgesg82:
## Re: boot-time fault clearing propagation targets

Valid catch. Fixed in `b7b229fd`.

**Root cause:** `reset_runtime_storage()` was clearing `propagation_targets`, but it's called by `reconstruct_runtime_machine()` which is invoked by `rebuild_as_fault()` during `request_fault()` when `!runtime_started`. This cleared targets before `propagate_fault()` ran.

**Fix:** Moved the propagation state reset out of `reset_runtime_storage()` and into:
- `install_runtime<>()` — resets targets on policy reinstall (after `reconstruct_runtime_machine`)
- `TestAccess::FaultController::clear()` — resets targets for test isolation

`reconstruct_runtime_machine()` (and by extension `rebuild_as_fault()`) no longer touches propagation state, so boot-time faults propagate correctly.

jorgesg82:
## Copilot Review Round 3 Responses

Addressed 2 of 5 new comments in `d781bb5b`:

| # | Comment | Action |
|---|---------|--------|
| 3 | `send_order()` return value ignored; emit diagnostic on failure | **Fixed** — `propagate_fault()` now checks the `send_order()` return and emits a `runtime_warning` diagnostic + `flush_urgent()` on failure. |
| 5 | Changeset summary says "suppressing re-propagation on received faults" — stale wording | **Fixed** — reworded to "with single-shot loop prevention via the `!faulted` guard". |
| 1 | `register_fault_propagation` should return `bool` for null/capacity failures | **Won't fix** — the only caller is init-time configuration in `VCU.hpp`. A return value would be ignored. If registration fails, the board simply doesn't propagate — which is a misconfiguration bug, not a runtime failure. |
| 2 | Validate `fault_order->get_id() == 0` and return `bool` | **Won't fix** — same reasoning as round 2: the API is intentionally generic. Coupling ST-LIB to a specific order ID would be a layering violation. |
| 4 | No unit tests (dup of earlier comment) | **Won't fix in this PR** — same as before: test suite doesn't build with `STLIB_ETH` defined. |

PR #677, opened 2026-07-13
fix: exti does not turn on automatically now

jorgecanut:
Mi primer commit, dadle amor


jorgesg82:
Would this affect to the EXTI used in other Boards? Because maybe that is been taking into account in other boards and you will brake them after pushing this. Please @victor-Lopez25 @FoniksFox confirm that this is fine for you both

